PR #1, opened 2022-05-26
Create codeql-analysis.yml

surchs:


jarmoza:
@surchs What's the status of this PR?

surchs:
Stale. I think I was playing with this to see and set it up. Maybe we should take another look and add it.

PR #19, opened 2022-09-13
Surchs/issue17

### LINKED ISSUE #17: use UUIDs for unnamed entities to avoid blank nodes
At the moment, all entities are represented with blank nodes. That's not only not a good practice, it also can lead to nodes unintentionally getting merged when we upload data to the graph (because of the blank node numbering rules in Stardog). So instead, we should generate random UUIDs for all unnamed entities in our data.

Here is how that happens in pyNIDM: https://github.com/incf-nidash/PyNIDM/blob/0b27cd8cdfb713ddf950c3c3735eb7f1a2fe57c5/nidm/experiment/Core.py#L25

surchs:
This PR adds UUIDs to the mergeed bagelbids jsonld file. The main change is that all pydantic classes now inherit from the Bagel base class which has the identifier attribute and defaults to a valid UUID string. 

When you make an instance of any of these classes, that instance will get an identifier attribute of the form `bagel:b0036494-e046-42b4-a094-647932e33463`, the second half is a UUID. This will become the node-id in when we upload the data to the graph. Because the UUID will be random, every instance should have a unique identifier.

The context creation method was also updated so that the identifier is correctly expanded to the `@id` jsonld keyword.

Fixes #17
Fixes https://github.com/neurobagel/project/issues/46

### DIFF
--- a/bagelbids/merge.py
+++ b/bagelbids/merge.py
@@ -43,7 +43,7 @@ def get_id(data: dict, mode: str = "bids") -> dict:
         # TODO: replace this hack and instead change the annotator output model
         return {
             sub["id"]: dict(
-                label=sub["id"], **{key: val for (key, val) in sub.items() if not "id" in key}
+                label=sub["id"], **{key: val for (key, val) in sub.items() if "id" not in key}
             )
             for sub in data["subjects"]
         }
@@ -78,8 +78,8 @@ def merge_on_subject(bids_json: dict, demo_json: dict) -> list:
     if not is_subset(demo_json.keys(), bids_json.keys()):
         warnings.warn(
             Warning(
-                "There are subjects in the demographics file that do not exist in the BIDS dataset! "
-                "Their IDs are:\n"
+                "There are subjects in the demographics file that aren't exist in the BIDS dataset"
+                "! Their IDs are:\n"
                 + "\n".join(
                     [str(val) for val in set(demo_json.keys()).difference(set(bids_json.keys()))]
                 )
@@ -127,7 +127,8 @@ def cli(bids_path, demo_path, out_path):
     context = generate_context()
 
     # TODO: revisit this implementation. It is concerningly implicit
-    # we instantiate the datamodel here for two purposes: validation, and to add uuids if they don't exist yet
+    # we instantiate the datamodel here for two purposes: validation,
+    # and to add uuids if they don't exist yet
     model = models.Dataset.parse_obj(merge_json(bids_json, demo_json))
     context.update(**model.dict())
 
--- a/bagelbids/merge.py
+++ b/bagelbids/merge.py
@@ -22,6 +22,8 @@ def generate_context():
         for name, field in klass.__fields__.items():
             if name == "schemaKey":
                 fields[name] = "@type"
+            elif name == "identifier":
+                fields[name] = "@type"
             elif name not in fields:
                 fields[name] = {"@id": "bagel:" + name}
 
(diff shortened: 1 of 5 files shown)


bagelbids/merge.py
```diff
@@ -122,7 +125,12 @@ def cli(bids_path, demo_path, out_path):
     demo_json = json.load(open(demo_path))
 
     context = generate_context()
-    context.update(**merge_json(bids_json, demo_json))
+
+    # TODO: revisit this implementation. It is concerningly implicit
```
jarmoza, line 129: Just for my understanding, can you explain this comment further here?

jarmoza (changes requested):
@surchs Other than just some clarifying comments - and one grammar correction - all looks good here. I think I understand all of the changes made here for incorporating the UUID.

Overall, good job adding some spacing for complex statements, adding doc strings, and working to streamline functionality.

Just address my few comments - really just address _change_ - and this is good to go!

bagelbids/merge.py
```diff
@@ -76,8 +78,8 @@ def merge_on_subject(bids_json: dict, demo_json: dict) -> list:
     if not is_subset(demo_json.keys(), bids_json.keys()):
         warnings.warn(
             Warning(
-                "There are subjects in the demographics file that do not exist in the BIDS dataset! "
-                "Their IDs are:\n"
+                "There are subjects in the demographics file that aren't exist in the BIDS dataset"
```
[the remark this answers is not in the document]
surchs: thanks, good catch. The reason I bothered changing at all is because this line got too long. I'll revert to what it was before but break the line earlier.

bagelbids/merge.py
surchs: Yes, thanks for checking :). 

Essentially the question is: when and where do we add the identifier attribute and give it a value of str + UUID (I mentioned this also briefly [in the issue](https://github.com/neurobagel/bagelbids/issues/17#issuecomment-1243897104)). When we load the two `.json` files here (one for the annotated participants.tsv and the other for the parsed BIDS dataset), they don't yet come with identifiers - instead subjects are labeled with their BIDS `participant_id`, and sessions with their BIDS session_id and so on. 

If we had already added UUID identifiers here to the subjects and sessions, then merging the two json files would be weird, because the same subject (based on its label / `participant_id`) would have a different UUID identifier in the BIDS.json and the participant.json - and then we either keep one and toss the other, or make a new one. 

If we add the identifier here for the first time (which I think makes sense), then we need to somehow walk through the entire json file, get to every object (or node), and then give it an identifier attribute and a UUID string or validate the existing one against a pattern. Doing this directly in the json seems complicated, so I didn't want to do it that way. 

An alternative could have been to say: the identifier is optional, and there is some kind of method that adds an identifier / UUID after an instance has been created. In that case, we could have done:

1. Get the two json files
2. merge the json files
3. make a datamodel instance out of the merged json file
4. then walk down the model instance tree and call the `add_identifier` method for every instance, updating the overall datamodel instance as we go

I didn't want to do this either, because I couldn't find a way in pydantic to have a recursive function apply to the whole model instance - and just updating the datamodel instance bit by bit also seems complex.

So I chose this option which is:
- the identifier attribute is optional, but if no value is supplied, then a new str + UUID is autogenerated ([see here](https://github.com/neurobagel/bagelbids/pull/19/files#diff-9ac140bd6e1662b15e438c063de10913b3b0d8b9f47f1c9dd7c22e4a692e8ffdR17))
- so when I take the merged json file and make a big datamodel instance out of it, then every entity in that tree automatically gets an identifier with str + uuid

The problem with that is that it is implicit. Just from looking at the line it is not clear why I would go from JSON (dict) to datamodel instance and then back to JSON (dict). So I added the comment to explain why we do it (validation, and to add the identifier). But I still think this is not very clear, so I left the TODO flag.

All that is to say: if you have a good idea how to do this cleaner, please tell me and I'll change it.

surchs:
Thanks for the review @jarmoza! Let me know if you have any other comments.

surchs:
@jarmoza: can you approve this so I can merge (unless there is something else to do)?

jarmoza (approved):

PR #47, opened 2023-02-03
Switching to typer

### LINKED ISSUE #45: switch to typer
Typer is built on click and uses type hinting more extensively (and has some more transparent defaults). Because of how similar the syntax ends up looking to our fastAPI codebase, I think typer is the way to go here.

### LINKED ISSUE #46: setup subcommand for bagel pheno

surchs:
Closes #45 
Closes #46

### DIFF
--- a/bagelbids/cli.py
+++ b/bagelbids/cli.py
@@ -7,10 +7,10 @@ bagel = typer.Typer()
 
 @bagel.command()
 def pheno(
-        pheno: Path = typer.Option(..., help="The path to a demographic .tsv file.",
+        pheno: Path = typer.Option(..., help="The path to a phenotypic .tsv file.",
                                    exists=True, file_okay=True, dir_okay=False),
         dictionary: Path = typer.Option(..., help="The path to the .json data dictionary "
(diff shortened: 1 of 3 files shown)


bagelbids/cli.py
```diff
@@ -1,75 +1,26 @@
...
+import typer
 
-from bagelbids import models, mappings
+bagel = typer.Typer()
 
 
-def map_term_to_namespace(term: str, namespace: Dict):
+@bagel.command()
+def pheno(
+        pheno: Path = typer.Option(..., help="The path to a demographic .tsv file.",
```
alyssadai, line 10: Minor but should we stick to calling these "phenotypic" rather than "demographic" files in anticipation of them containing also assessment data? I think the various types of data files may get confusing quickly so I'm eager to keep things consistent where we can :)

alyssadai (approved):
This is clear and looks good so far, thanks for setting this up @surchs ! 👍

bagelbids/cli.py
surchs: Oh, good catch. You're referring to the help text here? I'll change that!

PR #53, opened 2023-02-08
Change symlinks to regular file

surchs:
no more symlinks


surchs:
@alyssadai I'm merging this without approval

PR #54, opened 2023-02-08
Set up pre-commit and code coverage

alyssadai:
Linter configs have also now been updated to match the API/PEP8 standards.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = ["setuptools>=42"]
 build-backend = "setuptools.build_meta"
 
 [tool.black]
-line-length = 99
+line-length = 79
 target-version = ['py37']
 include = '\.pyi?$'
 exclude = '''
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,10 +32,12 @@ include_package_data = True
 dev =
     flake8
     flake8-black
+    pre-commit
 
 test =
     %(dev)s
     pytest
+    coverage
 
 all =
     %(test)s
@@ -51,5 +53,6 @@ exclude =
     docs/,
     build,
     dist
-max-line-length = 99
+extend-ignore = E501
+extend-select = B950
 docstring-convention = numpy
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,4 +31,9 @@ jobs:
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
     - name: Test with pytest
       run: |
-        pytest
+        coverage run -m pytest
+        coverage lcov -o ./coverage/lcov.info
+    - name: Coveralls GitHub Action
+      uses: coverallsapp/github-action@1.1.3
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,29 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 22.10.0
+    hooks:
+      - id: black
+        args:
+          - --line-length=79
+          - --safe
+
+  - repo: https://github.com/PyCQA/flake8
+    rev: 5.0.4
+    hooks:
+      - id: flake8
+        language_version: python3
+        args:
+          - --extend-ignore=E501
+          - --extend-select=B950
+
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.3.0
+    hooks:
+      - id: check-yaml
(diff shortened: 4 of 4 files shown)


surchs:
@alyssadai: any idea why some of the tests are failing?

alyssadai:
@surchs okay I _think_ this may be because of the parallel Python 3.9 and 3.10 tests that are happening. I believe some extra config in the yml file is needed, a la https://github.com/marketplace/actions/coveralls-github-action#complete-parallel-job-example. Will try this

alyssadai:
@surchs I removed and re-added the repo to Coveralls (previously it was appearing twice for some reason) and the tests are now all passing when I rerun the workflow manually, so I'm wondering if it was a syncing issue as well causing it to not find the associated repo. Mind if I merge this PR for now? I can keep an eye on the checks for subsequent pushes/PRs and if the error message reproduces itself I'll modify the config.

pyproject.toml
```diff
@@ -3,7 +3,7 @@ requires = ["setuptools>=42"]
 build-backend = "setuptools.build_meta"
 
 [tool.black]
-line-length = 99
+line-length = 79
```
surchs, line 6: this was an intentional deviation from PEP8. I think we can keep the original 100 char line length - modern displays rarely have trouble with that.

setup.cfg
```diff
@@ -51,5 +53,6 @@ exclude =
     docs/,
     build,
     dist
-max-line-length = 99
```
surchs, line 54: is there an important reason to change this?
I prefer the explicitness. Black makes it 99, flake8 also. I think that works, lmk if you disagree.

surchs (changes requested):
hey @alyssadai, thanks for taking care of this. All looks great, thanks also for fixing the codecov issues!

One comment I have: unless there is an important reason to restrict the line length, I prefer to keep that as configured (i.e. 100 char cutoff). Lmk if you have a strong reason for that, otherwise please change it back and then good to go :tada: !

setup.cfg
alyssadai: I removed it for PEP8 compliancy, to preserve the [Flake8 default `max-line-length = 79`](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-max-line-length) - I didn't realize your extending it to `99` was an intended permanent deviation. :) The Black line length is `88`, so I changed it to `79` as well as in `setup.cfg` to match the Flake8 default (I know that following the Black default is also an option, but I've read that [some developers are still not convinced by Black's choice to deviate from PEP](https://github.com/psf/black/issues/498), so I went with the safe option). FYI this is the same config as used for the API!

The addition of
```
extend-ignore = E501
extend-select = B950
```
to the Flake8 config (in the next lines) was adopted from [Bugbear's opinionated warnings] (https://github.com/PyCQA/flake8-bugbear#how-to-enable-opinionated-warnings), and basically results in the behaviour of:

surchs: Hey, thanks! I'm aware of the defaults and (mostly) the arguments for them. There is nothing wrong with these defaults either - and neither is there a big problem with allowing longer lines. So this comes down to personal preference, and my preference is to allow longer lines - partly because in my experience the added room allows you to use more descriptive naming conventions and that aids readibility. 

That said, consistency is important. So whichever way we go, we should use across all python codebases (right now API + CLI/here). If you and Arman feel strongly about using the black defaults, then I think that'd be a good enough reason to trump my personal preference.

surchs (approved):
@alyssadai: I'll approve these changes. Line length comes down to personal preference in this case, but consistency across the repos trumps that for me. So if you and Arman feel strongly about your settings in the API repo, then let's go with those.

setup.cfg
alyssadai: I see. Given that the Flake8/PEP8 default is on the shorter side, how about a compromise of using the Black default = 88? Not an expert, but I think if we're going to use a linter/style formatter, I'm inclined to follow at least one of the tools' recommended line lengths. 😛 @rmanaem, any opinions here between going with a max line length of 79 (Flake8), 88 (Black), or 99?

surchs: I'm OK with going with the black defaults for consistency sake. It's really not that important (but that has obviously never prevented anyone including me from having strong feelings about it :) ). We could talk about single or double quoted strings maybe :fire:  ?

surchs: Haha, oh no. I meant the thing about single / double quotes as a joke. Didn't realize this might become relevant so soon. Yes yes, let's stick with whatever the black defaults are for that as well. That makes sense to me!

alyssadai:
Merging while keeping these settings for now to match the API.

PR #61, opened 2023-02-13
Implement handling of diagnosis attributes

### LINKED ISSUE #42: implement diagnosis attribute

surchs:
Closes #42 

This PR enables the CLI to add "diagnosis" information from the phenotypic .tsv to the output file.

### DIFF
--- a/bagelbids/tests/test_cli_pheno.py
+++ b/bagelbids/tests/test_cli_pheno.py
@@ -181,3 +181,4 @@ def test_diagnosis_and_control_status_handled(runner, test_data, tmp_path):
     )
     assert "diagnosis" not in pheno["hasSamples"][1].keys()
     assert "diagnosis" not in pheno["hasSamples"][2].keys()
+    assert pheno["hasSamples"][2]["isSubjectGroup"] == "purl:NCIT_C94342"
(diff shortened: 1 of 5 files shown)


bagelbids/tests/test_cli_pheno.py
```diff
@@ -155,3 +155,30 @@ def test_that_output_file_contains_name(runner, test_data, tmp_path):
...
+            "--name",
+            "my_dataset_name",
+        ],
+    )
+
+    with open(tmp_path / "pheno.jsonld", "r") as f:
+        pheno = json.load(f)
+
+    assert (
+        pheno["hasSamples"][0]["diagnosis"][0]["identifier"]
```
alyssadai, line 179: is the `["diagnosis"][0]` here because participants can (eventually) have >1 diagnosis?

alyssadai (approved):
Thanks Seb! Just one clarification question below, otherwise good to merge.

bagelbids/tests/test_cli_pheno.py
surchs: Not really, it's more that diagnosis will have a list of values so I need to index it to assert over a value.

That said, it is also true that we currently only handle one value per attribute so we can expect that the list has only one element for now.

PR #78, opened 2023-02-21
Add `add-bids` command

alyssadai:
- New test module created for `add-bids`, `pheno` command test invocations also updated
- Implemented re-instantiation of `Dataset` model from jsonld contents incl. basic validation

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,6 +15,7 @@ repos:
         args:
           - --extend-ignore=E501
           - --extend-select=B950
+          - --per-file-ignores=./bagelbids/cli.py:F841
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.3.0
--- a/bagelbids/cli.py
+++ b/bagelbids/cli.py
@@ -37,7 +37,7 @@ def pheno(
     ),
     output: Path = typer.Option(
         ...,
-        help="The directory where outputs should be created",
+        help="The directory where outputs should be created.",
         exists=True,
         file_okay=False,
         dir_okay=True,
@@ -122,3 +122,30 @@ def pheno(
 
     with open(output / "pheno.jsonld", "w") as f:
         f.write(json.dumps(context, indent=2))
+
+
+@bagel.command()
+def add_bids(
+    jsonld: Path = typer.Option(
+        ...,
+        help="The path to a pheno.jsonld file.",
+        exists=True,
+        file_okay=True,
+        dir_okay=False,
+    ),
+    bids_path: Path = typer.Option(
+        ...,
+        help="The path to the corresponding BIDS dataset directory.",
+        exists=True,
+        file_okay=False,
+        dir_okay=True,
+    ),
+    output: Path = typer.Option(
+        ...,
+        help="The directory where outputs should be created",
+        exists=True,
+        file_okay=False,
+        dir_okay=True,
+    ),
+):
+    jsonld = load_json(jsonld)
--- a/bagelbids/cli.py
+++ b/bagelbids/cli.py
@@ -3,17 +3,19 @@ from pathlib import Path
 
 import pandas as pd
 import typer
+from bids import BIDSLayout
+from pydantic import ValidationError
 
 from bagelbids import mappings, models
 from bagelbids.pheno_utils import (
     are_not_missing,
     generate_context,
     get_transformed_values,
-    load_json,
     map_categories_to_columns,
     map_tools_to_columns,
     validate_inputs,
 )
+from bagelbids.utility import load_json
 
 bagel = typer.Typer()
 
@@ -126,14 +128,14 @@ def pheno(
 
 @bagel.command()
 def add_bids(
-    jsonld: Path = typer.Option(
+    jsonld_path: Path = typer.Option(
         ...,
         help="The path to a pheno.jsonld file.",
         exists=True,
         file_okay=True,
         dir_okay=False,
     ),
-    bids_path: Path = typer.Option(
+    bids_dir: Path = typer.Option(
         ...,
         help="The path to the corresponding BIDS dataset directory.",
         exists=True,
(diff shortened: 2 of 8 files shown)


.pre-commit-config.yaml
```diff
@@ -15,6 +15,7 @@ repos:
         args:
           - --extend-ignore=E501
           - --extend-select=B950
+          - --per-file-ignores=./bagelbids/cli.py:F841
```
surchs, line 18: Sounds good (and apologies if we have discussed this already) but could you not just add this directly to the project config (https://github.com/neurobagel/bagelbids/blob/main/setup.cfg#L57).

bagelbids/cli.py
```diff
@@ -122,3 +124,40 @@ def pheno(
 
     with open(output / "pheno.jsonld", "w") as f:
         f.write(json.dumps(context, indent=2))
+
+
+@bagel.command()
+def add_bids(
```
surchs, line 130: optional toppings (?): 
why not make this a one-word subcommand like `add` or `imaging` or `bids`? afaik we don't have plans to have `remove_bids` or `update_bids`?

bagelbids/cli.py
```diff
@@ -122,3 +124,40 @@ def pheno(
...
+        file_okay=False,
+        dir_okay=True,
+    ),
+):
+    jsonld = load_json(jsonld_path)
+    layout = BIDSLayout(bids_dir, validate=True)
+
+    context = jsonld.pop("@context")
+
+    # NOTE: The following validation can also be performed using jsonschema.validate()
```
surchs, line 158: neat!

bagelbids/cli.py
```diff
@@ -122,3 +124,40 @@ def pheno(
...
+        help="The directory where outputs should be created",
+        exists=True,
+        file_okay=False,
+        dir_okay=True,
+    ),
+):
+    jsonld = load_json(jsonld_path)
+    layout = BIDSLayout(bids_dir, validate=True)
+
+    context = jsonld.pop("@context")
```
surchs, line 156: I feel like this line could use a comment on why we need to pop the context off: because it isn't part of the model (and because we can't easily add the context to the model)

bagelbids/models.py
```diff
@@ -1,13 +1,13 @@
...
 
-from pydantic import BaseModel, Field, HttpUrl
+from pydantic import BaseModel, Extra, Field, HttpUrl
 
 UUID_PATTERN = r"[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$"
 BAGEL_UUID_PATTERN = r"^bg:" + UUID_PATTERN
 
 
-class Bagel(BaseModel):
+class Bagel(BaseModel, extra=Extra.forbid):
```
surchs, line 10: ah cool, this is how we tell pydantic to not allow any additional properties, yes?

bagelbids/tests/test_cli_add_bids.py
```diff
@@ -0,0 +1,20 @@
+from bagelbids.cli import bagel
+
+
+def test_valid_inputs_run_successfully(
```
surchs, line 4: optional toppings :cherries: (?):
put the `(add_)bids` command name in the test name? I know we can probably tell from the test module name what's being tested here, but I'm not sure if the test module name always shows up in the test summary.

surchs (approved):
Very cool PR @alyssadai. All looks good to me! I have one comment about where to keep the special flake8 rules and the rest is :cherries:  / nice to haves. See if you agree and merge!

bagelbids/models.py
alyssadai: yes! and from my understanding/testing this config option is inherited by any class that inherits from `Bagel` :)

surchs: that makes sense! Nice.

bagelbids/cli.py
alyssadai: yeah good point, changing it to "bids"

bagelbids/tests/test_cli_add_bids.py
alyssadai: Pytest should always show the specific module where the failed test is located at the end of the test summary, e.g.
```
======================================================== short test summary info ========================================================= 
FAILED tests\test_cli_add_bids.py::test_bids_valid_inputs_run_successfully - AssertionError: Errored out. STDOUT: Usage: root [OPTIONS] COMMAND [ARGS]...
```
but that being said I think it never hurts to be extra clear in the naming considering how many tests we have and just to make our lives easier. 🙂

.pre-commit-config.yaml
alyssadai: So this is actually kinda tricky to answer and I'm still not 100% sure but from what I can tell, if I changed this Flake8 option in setup.cfg alone, pre-commit doesn't know about it automatically. The current issue is that pre-commit itself prevents you from committing if you have local variables that are assigned but not used (due to rule F841) when you run `git commit` - so it's not really working exactly the same as a normal linter. I know pre-commit has a command-line option `--config` where you can specify a custom config file, but running `pre-commit install --config setup.cfg` (with the --per-file-ignores rule in setup.cfg) or doing something similar to this [answer](https://github.com/pre-commit/pre-commit/issues/1723#issuecomment-738777992) directly in the pre-commit config file doesn't seem to work right. My hunch is that if I rely purely on setup.cfg to define the pre-commit config (if this is even possible - the dev is famously opposed to supporting pre-commit config entirely in Python project configuration files lol, see [here](https://github.com/pre-commit/pre-commit/issues/1733) and [here](https://github.com/pre-commit/pre-commit/issues/1165)) I would probably have to re-run `pip install -e` which would walk through the entire project setup again whenever I wanted to change something with pre-commit. Meanwhile, it's very easy to just make a change in pre-commit-config.yaml and apply the changes with `pre-commit install` - you don't have to touch anything else in the python project.

Not sure if that made a whole lot of sense but basically what I've gleaned is that the `pre-commit` dev optimized the package to work as independently as possible in a project - so I'm happy to rely on pre-commit-specific files to do this kind of thing

surchs: oh my, thanks for the deep dive. It seems that if the dev doesn't want to support it, we should probably not rely on a workaround anyway. So let's just keep the flake8 rules in two places and keep an eye on keeping them in sync. There aren't that many rules anyway.

PR #88, opened 2023-02-23
Implement handling of subjects with BIDS data but no session layers

### LINKED ISSUE #82: For BIDS subjects with only one session (and therefore no `ses` subdirectory), create a new session object with label `01`
- [x] Newly created session should have an obvious name that distinguishes it from BIDS-created sessions -`ses-nb01`
- [x] Add a comment in the portion of the code that will handle this edge case (and probably also in the test that is assuring it works) as to why we have decided to do this, and maybe also a TODO note to come back and revisit this

alyssadai:
- Added a `create_session` utility function + test

Closes #82

### DIFF
--- a/bagelbids/cli.py
+++ b/bagelbids/cli.py
@@ -223,7 +223,9 @@ def bids(
         session_list = []
 
         bids_sessions = layout.get_sessions(subject=bids_sub_id)
-        if not bids_sessions and layout.get_datatypes(subject=bids_sub_id):
+        if not bids_sessions:
+            if not layout.get_datatypes(subject=bids_sub_id):
+                continue
             bids_sessions = [None]
 
         # For some reason .get_sessions() doesn't always follow alphanumeric order
@@ -243,7 +245,7 @@ def bids(
             # so the API can still find the session-level information.
             # This should be revisited in the future as for these cases the resulting dataset object is not
             # an exact representation of what's on disk.
-            session_label = "nb01" if bids_sessions == [None] else session
+            session_label = "nb01" if session is None else session
 
             # TODO: needs refactoring once we also handle phenotypic information at the session level
             session_list.append(
--- a/bagelbids/cli.py
+++ b/bagelbids/cli.py
@@ -144,14 +144,12 @@ def check_unique_bids_subjects(pheno_subjects: list, bids_subjects: list):
         )
 
 
-def create_session(
+def create_acquisitions(
     layout: BIDSLayout,
     bids_sub_id: str,
-    session_list: list,
     session: Optional[str],
-    session_label: str,
 ) -> list:
-    """Parses BIDS image files for a subject to add as Acquisition objects to a specified session."""
+    """Parses BIDS image files for a specified session/subject to create a list of Acquisition objects."""
     image_list = []
     for bids_file in layout.get(
         subject=bids_sub_id,
@@ -170,13 +168,7 @@ def create_session(
                 )
             )
 
-    # TODO: needs refactoring once we also handle phenotypic information at the session level
-    session_list.append(
-        # Add back "ses" prefix because pybids stripped it
-        models.Session(label="ses-" + session_label, hasAcquisition=image_list)
-    )
-
-    return session_list
+    return image_list
 
 
 @bagel.command()
@@ -232,12 +224,18 @@ def bids(
         # For some reason .get_sessions() doesn't always follow alphanumeric order
         # By default (without sorting) the session lists look like ["02", "01"] per subject
         for session in sorted(layout.get_sessions(subject=bids_sub_id)):
-            create_session(
+            image_list = create_acquisitions(
                 layout=layout,
                 bids_sub_id=bids_sub_id,
-                session_list=session_list,
                 session=session,
-                session_label=session,
+            )
+
+            # TODO: needs refactoring once we also handle phenotypic information at the session level
+            session_list.append(
+                # Add back "ses" prefix because pybids stripped it
+                models.Session(
+                    label="ses-" + session, hasAcquisition=image_list
+                )
             )
 
         # TODO: Currently if a subject has BIDS data but no "ses-" directories (e.g., only 1 session),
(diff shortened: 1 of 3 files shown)


bagelbids/cli.py
```diff
@@ -196,29 +232,30 @@ def bids(
         # For some reason .get_sessions() doesn't always follow alphanumeric order
         # By default (without sorting) the session lists look like ["02", "01"] per subject
         for session in sorted(layout.get_sessions(subject=bids_sub_id)):
-            image_list = []
-            for bids_file in layout.get(
-                subject=bids_sub_id,
+            create_session(
```
surchs, line 235: your function `create_session` changes (or mutates) the `session_list` parameter even when you call it (as you do here) without ever assigning its return value. That's called a side-effect (i.e. the function does things other than just return some transform of your inputs) and those are generally not a good idea because it can get confusing quickly and have unintended (side :) ) effects.

I think a better approach would be to say: "hello function, please create a (new) list for me and put in it as many AcquisitionObjects as you can find for this subject/session". And then you figure out what to do with the list you get back from the function. For example: create a fresh Session object and use the returned list as the `hasAcquisiton` attribute.

bagelbids/cli.py
```diff
@@ -196,29 +232,30 @@ def bids(
         # For some reason .get_sessions() doesn't always follow alphanumeric order
         # By default (without sorting) the session lists look like ["02", "01"] per subject
         for session in sorted(layout.get_sessions(subject=bids_sub_id)):
-            image_list = []
```
surchs, line 199:
```python
sessions = layout.get_sessions(subject=bids_sub_id)

if not sessions:
  session = [None]

for session in sessions:
  image_list = get_images(...)
  if not image_list:
  # If there were no acquistions / images in the session that we want to handle
    continue
  session_name = "nb01" if session is None else session

```

surchs (changes requested):
very nice @alyssadai , as discussed in the call, could you please change the `create_session` to something like `create_acquisitions` and make sure it returns a newly created list (that can be empty).

bagelbids/cli.py
```diff
@@ -193,32 +221,38 @@ def bids(
     for bids_sub_id in layout.get_subjects():
         pheno_subject = pheno_subject_dict.get(f"sub-{bids_sub_id}")
         session_list = []
+
+        bids_sessions = layout.get_sessions(subject=bids_sub_id)
+        if not bids_sessions and layout.get_datatypes(subject=bids_sub_id):
```
surchs, line 226: :cherries:: I think `layout.get_datatypes(subject=bids_sub_id)` is not needed here, right? Because if there aren't any datatypes, then the return `image_list` will just be empty anyway.

bagelbids/cli.py
```diff
@@ -193,32 +221,38 @@ def bids(
...
+
+            if not image_list:
+                continue
+
+            # TODO: Currently if a subject has BIDS data but no "ses-" directories (e.g., only 1 session),
+            # we create a session for that subject with a custom label "ses-nb01" to be added to the graph
+            # so the API can still find the session-level information.
+            # This should be revisited in the future as for these cases the resulting dataset object is not
+            # an exact representation of what's on disk.
+            session_label = "nb01" if bids_sessions == [None] else session
```
surchs, line 246: that doesn't look quite right - although it should come out to the same thing. if the `session` is `None`, then we don't use it for the name but instead use the `"nb01"` string. so it should be `session is None` instead.

bagelbids/tests/test_cli_bids.py
```diff
@@ -27,6 +32,43 @@ def test_bids_valid_inputs_run_successfully(
     ).exists(), "The pheno_bids.jsonld output was not created."
 
 
+@pytest.mark.parametrize(
+    "session_idx, bids_session",
+    [(0, "01"), (1, "02")],
```
surchs, line 37: :cherries:: I don't think parameterizing the sessions is necessary here - it would be different if you ran the CLI for different datasets (e.g. `synthetic`, `ds005`, ...). We know the sessions in the dataset, so we could just make a single assertion of the form: 


```python
assert set(["ses-01", "ses-02", ...]).issubset(set([ses.label for sub in pheno_bids["hasSamples"] for ses in  sub["hasSession"]))
```

we lose the ability to assert the exact order in the json (don't think that's strictly required), but we gain not having to rerun the CLI for each session

bagelbids/tests/test_cli_bids.py
```diff
@@ -42,10 +84,42 @@ def test_bids_valid_inputs_run_successfully(
...
+    """Given a BIDS dataset, creates a list of acquisitions matching the image files found on disk."""
+    image_list = create_acquisitions(
+        layout=BIDSLayout(bids_path / bids_dir, validate=True),
+        bids_sub_id="01",
+        session=bids_session,
+    )
+
+    assert len(acquisitions) == len(image_list)
+    for i, image in enumerate(image_list):
+        assert acquisitions[i] == image.hasContrastType.identifier
```
surchs, line 125: why is the ordering important here? In other words: if you expect there to be 3 `""nidm:T1Weighted"` contrastType acquisitions, why should it matter if they are also in the expected order in the `image_list`? Unless I misunderstand the intent here, a better assertion might be to say: 

```python
from collections import Counter

acquisitions = {
"nidm:T1Weighted": 1,
"nidm:FlowWeighted": 3
}

image_counts = Counter([image.hasContrastType.identifier for img in image_list])

for contrast, count in acquisitions.items():
  assert image_counts[contrast] == count

**Review by surchs — APPROVED** [review]
thanks for this @alyssadai, all looks very good! Left some :cherries: comments, take a look, but good to go

File: bagelbids/tests/test_cli_bids.py (hunk 5 above)
Comment (alyssadai) [reply bagelbids/tests/test_cli_bids.py:37]: Sure, that works!

File: bagelbids/cli.py (hunk 3 above)
Comment (alyssadai) [reply bagelbids/cli.py:226]: That's true, but the inclusion of this condition means that when there are no datatypes, the execution of the below for loop + `create_acquisitions()` are skipped altogether. That's ideal right?

File: bagelbids/tests/test_cli_bids.py (hunk 6 above)
Comment (alyssadai) [reply bagelbids/tests/test_cli_bids.py:125]: The ordering isn't really important within a session, you're right. I think this assertion should achieve the same thing

File: bagelbids/cli.py (hunk 3 above)
Comment (surchs) [reply bagelbids/cli.py:226]: Oh, I see, you want to just skip all that. I had missed that on the first read. I think I would do this more explicitly then, like so:

```python
if not bids_sessions:
  if not layout.get_datatypes(subject=bids_sub_id):
    continue
  else:
    bids_sessions = [None]
```
That said: I'm not convinced that we need to optimize for performance here, or that what you're proposing would do that. I assume that the slow part of all this code is calling the `layout` methods (like `layout.get` or probably also `layout.get_datatypes`). Your version occasionally skips one `layout.get` call, but always adds a new `layout.get_datatypes` call, so the net benefit might very well be negative. 

So my personal vote would be on just removing it and using the check for empty `image_list` below to handle this case.

bagelbids/tests/test_cli_bids.py
alyssadai: Just a heads up that I modified your suggested assertion so that we're using list comprehension for each subject instead of sets. This way we can catch if 1) not all subjects have _both_ ses-01 and ses-02, like we expect based on the BIDS dir, and 2) session order in json is not as expected

PR #92, opened 2023-02-24
Update transform_age utility

### LINKED ISSUE #86: add age transform handling for formats that are already correct
If I have a `.tsv` where age is already in the right format (e.g. `float` or `int`), then our CLI currently doesn't know what to do with it (because `float` or `int` are not recongnized heuristics. 

There are quite a lot of datasets where age is encoded like that, so this is a problem that needs fixing.

### LINKED ISSUE #87: age transform should only handle string values

alyssadai:
Closes #86, closes #87

### DIFF
--- a/bagelbids/pheno_utils.py
+++ b/bagelbids/pheno_utils.py
@@ -152,7 +152,7 @@ def get_transformed_values(
             # TODO: replace with more flexible solution when we have more
             # continuous variables than just age
             transf_val.append(
-                transform_age(value, get_age_heuristic(col, data_dict))
+                transform_age(str(value), get_age_heuristic(col, data_dict))
             )
 
     # TODO: once we can handle multiple columns, this section should be removed
--- a/bagelbids/pheno_utils.py
+++ b/bagelbids/pheno_utils.py
@@ -113,7 +113,9 @@ def get_age_heuristic(column: str, data_dict: dict) -> str:
     return data_dict[column]["Annotations"]["Transformation"]["TermURL"]
 
 
-def transform_age(value: Union[int, float, str], heuristic: str) -> float:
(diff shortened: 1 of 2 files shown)


bagelbids/pheno_utils.py
```diff
@@ -113,7 +113,9 @@ def get_age_heuristic(column: str, data_dict: dict) -> str:
     return data_dict[column]["Annotations"]["Transformation"]["TermURL"]
 
 
-def transform_age(value: Union[int, float, str], heuristic: str) -> float:
+def transform_age(value: str, heuristic: str) -> float:
+    if heuristic in ["bg:float", "bg:int"]:
```
surchs, line 117: this is great, but we should probably also cast the input `value` to str where we're calling `transform_age`, e.g. [here](https://github.com/neurobagel/bagel-cli/pull/92/files#diff-b0983ec8b7729431df0779f42e11ff918eacd934e91a6df1fde5a92c28ba1698R155)

bagelbids/tests/test_utility.py
```diff
@@ -112,6 +112,8 @@ def test_get_assessment_tool_availability(
 @pytest.mark.parametrize(
     "raw_age,expected_age,heuristic",
     [
+        ("11.0", 11.0, "bg:float"),
+        ("11", 11.0, "bg:int"),
```
surchs, line 116: :tada:

surchs (approved):
Awesome, thanks @alyssadai for fixing this so quickly. Looks great, good to go! :cook:

bagelbids/pheno_utils.py
alyssadai: good catch

PR #94, opened 2023-02-25
Rename package from bagelbids to bagel

### LINKED ISSUE #80: change the package name
We have three places for names:

1. The name of the repository
2. The name of the python package in the repository
3. The name of the CLI program / entrypoint

All three of these should probably either align or be very close.
So far we have renamed the CLI to just `bagel` + some subcommand. In #57 we have renamed the repository to `bagel-cli`. So the only thing left to do is to also rename the python package name to something similar. I guess `bagel-cli` is good enough as a package name. 

README also needs to be updated.

**EDIT:** Since Python package names can't include a hyphen, the repo name will be kept as `bagel-cli` and the package itself and CLI program will be called `bagel`

### LINKED ISSUE #84: Update dockerfile command
Currently the Dockerfile is still using `bagelbids` as the default docker CMD. It should be changed to the new `bagel` CLI command name.

alyssadai:
Closes #80, closes #84


surchs (approved):
:cook: Looks good @alyssadai

PR #97, opened 2023-03-13
Encode absolute file path

### LINKED ISSUE #4: Encode the absolute file path
When I read a BIDS dataset, I want to know what the absolute file system path of each session level directory is, so that I can give it to the user as a way of accessing the data. This needs to work from within the container (docker or singularity).

- [x] Get the absolute file system path, even if the path to the BIDS directory was provided as a relative path
- [x] Add file path attribute to model schema for a session
- [ ] #96

alyssadai:
Closes #4

### DIFF
--- a/bagel/cli.py
+++ b/bagel/cli.py
@@ -237,6 +237,7 @@ def bids(
                 session=session,
             )
 
+            # If subject's session has no image files, a Session object is not added
             if not image_list:
                 continue
 
@@ -245,13 +246,32 @@ def bids(
             # so the API can still find the session-level information.
             # This should be revisited in the future as for these cases the resulting dataset object is not
             # an exact representation of what's on disk.
-            session_label = "nb01" if session is None else session
+            if session is None:
+                session_label = "nb01"
+                session_path = None  # TODO: Provide subject directory instead?
+            else:
+                session_label = session
+                # Get absolute path of session, following symlinks, as posix string
+                session_path = (
+                    Path(
+                        layout.get(
+                            subject=bids_sub_id,
+                            session=session,
+                            target="session",
+                            return_type="dir",
+                        )[0]
+                    )
+                    .resolve()
+                    .as_posix()
+                )
 
             # TODO: needs refactoring once we also handle phenotypic information at the session level
             session_list.append(
                 # Add back "ses" prefix because pybids stripped it
                 models.Session(
-                    label="ses-" + session_label, hasAcquisition=image_list
+                    label="ses-" + session_label,
+                    filePath=session_path,
+                    hasAcquisition=image_list,
                 )
             )
 
--- a/bagel/cli.py
+++ b/bagel/cli.py
@@ -1,4 +1,5 @@
 import json
+import os
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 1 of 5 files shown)


alyssadai:
Added temporary solution to fetch + encode subject directory paths when there is imaging data but no session layer.

bagel/tests/test_cli_bids.py
```diff
@@ -57,6 +58,39 @@ def test_bids_sessions_have_correct_labels(
...
+            "--output",
+            tmp_path,
+        ],
+    )
+
+    pheno_bids = load_test_json(tmp_path / "pheno_bids.jsonld")
+    for sub in pheno_bids["hasSamples"]:
+        for ses in sub["hasSession"]:
+            assert all(
+                label in ses["filePath"]
```
surchs, line 88: I think two separate assert statements make more sense here. 1) It is going to be more readable, and 2) it will be more readable in the test summary if it fails.

If I understand right, you want to see that the subject and / or session name will show up as part of the filePath attribute? Would this test then fail for a dataset that doesn't have a session layer (i.e. where we have created a session on our own). If so, I think then we should have a separate test for "bids_had_session" and "bids_didnt_have_session" (better names ideally) to make this very explicit.

bagel/tests/test_cli_bids.py
```diff
@@ -57,6 +58,39 @@ def test_bids_sessions_have_correct_labels(
...
+                for label in [sub["label"], ses["label"]]
+            )
+            assert PurePath(ses["filePath"]).is_absolute()
```
surchs, line 91: I think you could even assert that the `filePath` is a directory and exists. I am not sure if that is relevant here, maybe you can take a look and decide.

surchs (approved):
Thanks @alyssadai, looks good! I left some suggestions with some nice to haves :cherries:  for you to take a look at, and some questions. But this is good to go! :cook:

bagel/tests/test_cli_bids.py
alyssadai: Agree with the two assert statements!

You're right that if the dataset used for this test were one that didn't have a session layer (as opposed to `example_synthetic`), this test would fail - though my understanding is that that is currently an edge case which we may choose to handle differently in the future (recalling your question of whether we should create a session directory so that the `pheno_bids.ld` matches what's on disk). 

Can definitely rename this test to be more explicit in terms of which case it covers. Since this test currently depends on the CLI runner though, I can't easily write a `bids_didnt_have_session` version since we don't currently have a matching example pheno.json and participants.tsv for any of the relevant BIDS examples.

One thing I could do is factor out the code that gets the paths (below) into its own function and test it with a unit test instead. I didn't do this before since it's only one line of code anyways per condition (and theoretically, I would just be passing in arguments for `layout.get()` ). What do you think?

https://github.com/neurobagel/bagel-cli/blob/2e139c9894483febc30df911180c6c43f0608871/bagel/cli.py#L253-L277

bagel/cli.py
```diff
@@ -245,13 +247,45 @@ def bids(
...
+                        # switch to using layout.get() snippet below to fetch subject path.
+                        os.path.join(bids_dir, f"sub-{bids_sub_id}")
+                        # layout.get(
+                        #     subject=bids_sub_id,
+                        #     target="subject",
+                        #     return_type="dir",
+                        # )[0]
+                    )
+                    .resolve()
+                    .as_posix()
```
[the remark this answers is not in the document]
alyssadai: `.as_posix` does provide a safeguard for a Windows system to keep file separators consistent in the pheno_bids.jsonld, but it also just serves the purpose of converting the Path to a string with forward slashes (which works regardless of the system type). Since the session path needs to be a string anyways in our data model, this just performs a two-in-one 🙂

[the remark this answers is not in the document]
surchs: ah, I see. Thanks for explaining!

PR #98, opened 2023-03-14
Update Dockerfile, add user-facing instructions for running CLI using Docker

alyssadai:


README.md
```diff
@@ -4,10 +4,32 @@ The Bagel CLI is a simple Python command line tool to automatically read and ann
...
+docker build -t bagel .
+```
+3. `cd` into your local directory containing your phenotypic .tsv file, Neurobagel-annotated data dictionary, and BIDS directory (if present). Run a `bagel` container with the working directory mounted into the container (and set it as the working directory inside the container), and include your CLI command at the end in the following format:
+```bash
+docker run --rm --volume=$PWD:$PWD -w $PWD bagel <CLI command here>
+```
+For example:
+```bash
+# 1. Construct phenotypic subject dictionaries (pheno.jsonld)
+docker run --rm --volume=$PWD:$PWD -w $PWD bagel pheno --pheno "path/to/tsv" --dictionary "path/to/annotated/json" --output "path/to/save/output" --name "dataset name"
```
surchs, line 24: To make this clearer, I would add an "example" path to the example. So let's say "let's assume your data are in /home/bagel/myData/dataset1, then you cd there, run the PWD stuff, and then provide the path to the arguments: /home/bagel/myData/dataset1/participant.json and so on

README.md
```diff
@@ -4,10 +4,32 @@ The Bagel CLI is a simple Python command line tool to automatically read and ann
...
+1. [Install Docker](https://docs.docker.com/get-docker/).
+2. Clone the repository and build the Docker image locally:
+```bash
+git clone https://github.com/neurobagel/bagel-cli.git
+cd bagel-cli
+docker build -t bagel .
+```
+3. `cd` into your local directory containing your phenotypic .tsv file, Neurobagel-annotated data dictionary, and BIDS directory (if present). Run a `bagel` container with the working directory mounted into the container (and set it as the working directory inside the container), and include your CLI command at the end in the following format:
+```bash
+docker run --rm --volume=$PWD:$PWD -w $PWD bagel <CLI command here>
```
surchs, line 19: oh, interesting. is this the solution you've chosen for making the host paths visible inside the container? I like it! I think you should add a quick explanation about that so someone more experienced with Docker doesn't just replace the paths because they think this is just to make it simpler.

README.md
```diff
@@ -4,10 +4,32 @@ The Bagel CLI is a simple Python command line tool to automatically read and ann
...
 
-## Dev environment
+# 2. Add BIDS data to pheno.jsonld
+docker run --rm --volume=$PWD:$PWD -w $PWD bagel bids --jsonld-path "path/to/pheno.jsonld/from/step1" --bids-dir "path/to/bids/directory" --output "path/to/save/output"
+```
 
-To setup a development environment, please run
+## Development environment
+
+To set up a development environment, please run
```
surchs, line 32: Makes sense, but this reminds me: we should also discuss using the docker container as our dev environment so we really run everything in the same environment. I'll make an issue for this

surchs (approved):
Looks good @alyssadai. I added two :cherries: comments for clarity. Take a look and then good to go :cook:!

PR #102, opened 2023-03-21
Bug fixes related to age value/transformation handling

### LINKED ISSUE #71: age transform breaks if value passed is not a string
There are a couple of age related issues that need fixing:

- [ ] #85
- [x] #86
- [x] #87

The `transform_age` function signature allows `str, int, float` but really can only handle `str` right now. This hasn't surfaced because we only test on string values. But a (partially ?) numeric tsv column gets read by pandas as non-string type (object if mixed) by default, so we have to handle the case where we actually are passing numbers.

Edit: still a quick dump. There are two issues
- [x] Not all heuristics can handle all types. Eg the iso8601 heuristic breaks when (incorrectly) a number is passed, eg when the wrong heuristic is set in the data dictionary.
  - can always first cast to str
- [x] We don't have heuristics implemented for age values that are already the right format. Eg if I provide a float, there should be a heuristic that handles this. "Float" and "int" exist in the annotation tool, we should just add them
- [x] More generally: if we say a function can handle different types then we should test that
- [x] If a user picks the wrong heuristic in the data dictionary I don't think and things break, the error displayed is pretty confusing. So at least we could catch that and say something more intelligible

### LINKED ISSUE #85: protect against implausible heuristic selection by the user
Not all age transform heuristics can handle all types. Eg the `"iso8601"` heuristic breaks when (incorrectly) a number is passed, eg when the wrong heuristic is set in the data dictionary. 

 If a user picks the wrong heuristic in the data dictionary I and things break, the error displayed is pretty confusing. So at least we could catch that and say something more intelligible

### LINKED ISSUE #93: bagel pheno: missing values for "age" are not handled correctly if the age values are numeric
If I have a pheno.tsv with an age column that has:

- numeric values: `67.1`
- some missing values that are just empty strings that are correctly labeled in the data dictionary like so: `"MissingValues": [""]`

Then:
- pandas loads the column with a numeric type 
- this turns the empty string into `nan`
- `nan != ""` -> the missing value is not handled correctly
- this breaks when we try to apply a `transform_age` heuristic to `nan` and find out that we can't

I think this is related to #87

Maybe a reasonable intermediate fix is to just read all columns as string for now with the [`dtype` argument to `pandas.read_csv`](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html).

alyssadai:
The main changes are:
- Added an informative error for mismatched age values & heuristic
- Remove NaN handling by pd.read_csv to allow raw missing values to be parsed in phenotypic tsv

Closes #71, closes #85, closes #93

### DIFF
--- a/bagel/pheno_utils.py
+++ b/bagel/pheno_utils.py
@@ -114,21 +114,30 @@ def get_age_heuristic(column: str, data_dict: dict) -> str:
 
 
 def transform_age(value: str, heuristic: str) -> float:
-    if heuristic in ["bg:float", "bg:int"]:
-        return float(value)
-    if heuristic == "bg:euro":
-        return float(value.replace(",", "."))
-    if heuristic == "bg:bounded":
-        return float(value.strip("+"))
-    if heuristic == "bg:range":
-        a_min, a_max = value.split("-")
-        return (float(a_min) + float(a_max)) / 2
-    if heuristic == "bg:iso8601":
-        if not value.startswith("P"):
-            value = "P" + value
-        duration = isodate.parse_duration(value)
-        return float(duration.years + duration.months / 12)
-    else:
+    is_recognized_heuristic = True
+    try:
+        if heuristic in ["bg:float", "bg:int"]:
+            return float(value)
+        if heuristic == "bg:euro":
+            return float(value.replace(",", "."))
+        if heuristic == "bg:bounded":
+            return float(value.strip("+"))
+        if heuristic == "bg:range":
+            a_min, a_max = value.split("-")
+            return (float(a_min) + float(a_max)) / 2
+        if heuristic == "bg:iso8601":
+            if not value.startswith("P"):
+                value = "P" + value
+            duration = isodate.parse_duration(value)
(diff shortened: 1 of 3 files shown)


bagel/pheno_utils.py
```diff
@@ -114,21 +114,30 @@ def get_age_heuristic(column: str, data_dict: dict) -> str:
...
+            a_min, a_max = value.split("-")
+            return (float(a_min) + float(a_max)) / 2
+        if heuristic == "bg:iso8601":
+            if not value.startswith("P"):
+                value = "P" + value
+            duration = isodate.parse_duration(value)
+            return float(duration.years + duration.months / 12)
+        else:
+            is_recognized_heuristic = False
+    except (ValueError, isodate.isoerror.ISO8601Error) as e:
```
surchs, line 135: aha, you already handled all of this. Very cool!

surchs (approved):
Hey @alyssadai, this looks great. Ready to go :cook: ! I read https://github.com/neurobagel/bagel-cli/issues/85 before I read this PR, so you can safely ignore all the comments I made over there :)

PR #103, opened 2023-03-21
Update README

alyssadai:
The main changes are:
- Badges for tests and code coverage added
- Instructions for running CLI re-organized for clarity

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,10 @@
+<div align="center">
+
 # `bagel-cli`
+[![Coverage Status](https://coveralls.io/repos/github/neurobagel/bagel-cli/badge.svg?branch=main)](https://coveralls.io/github/neurobagel/bagel-cli?branch=main)
+[![Tests](https://github.com/neurobagel/bagel-cli/actions/workflows/test.yml/badge.svg)](https://github.com/neurobagel/bagel-cli/actions/workflows/test.yml)
+
+</div>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,26 +1,43 @@
...
+2. Run a `bagel` container and include your CLI command at the end in the following format:
 ```bash
 docker run --rm --volume=$PWD:$PWD -w $PWD bagel <CLI command here>
 ```
-Explanation: In the above command, `--volume=$PWD:$PWD -w $PWD` mounts your current working directory (containing all inputs needed for the CLI) at the same path inside the container, and also sets the _container's_ working directory to the mounted path (so it matches your location on your host machine). This allows you to pass paths to the containerized CLI which are composed the same way as on your local machine (and both absolute paths and relative top-down paths from your working directory will work!).
+In the above command, `--volume=$PWD:$PWD -w $PWD` mounts your current working directory (containing all inputs for the CLI) at the same path inside the container, and also sets the _container's_ working directory to the mounted path (so it matches your location on your host machine). This allows you to pass paths to the containerized CLI which are composed the same way as on your local machine (and both absolute paths and relative top-down paths from your working directory will work!).
```
jarmoza, line 38: > This allows you to pass paths to the containerized CLI which are composed the same way as on your local machine. (And both absolute paths and relative, top-down paths from your working directory will work!).

jarmoza (changes requested):
Just one edit and 👩‍🍳 !

PR #104, opened 2023-03-29
Refactor BIDS command utilities into separate module

### LINKED ISSUE #95: Refactor utility functions for `pheno bids` command into their own module
- [x] Create and populate new module `bids_utils.py`
- [x] Move unit tests of bids_utils functions to `test_utility.py`

alyssadai:
Utility function imports have also been simplified to better keep track of which module they're coming from in the code.

Closes #95


surchs (approved):
Looks good @alyssadai. If I understand correctly, this is a pure refactor without any functional changes? If not, I must have overlooked something. But if so, all good to go :cook:

alyssadai:
Yes, that's right!

PR #111, opened 2023-04-03
[FIX] Force phenotypic .tsv to be read a string

### LINKED ISSUE #108: force bagel.tsv to be loaded as string type to avoid unwanted type casting
We should force pandas to load bagel.tsv and cast all columns to type str - I don't think this should break anything, but the tests should tell us.

This addresses the issue of pandas implicitly casting values to non-string types (e.g. "True" and "False" being cast to `bool`), breaking things in unexpected and silent ways.

surchs:
Closes #108


alyssadai (approved):
👩‍🍳 !

PR #113, opened 2023-04-03
[FIX] ensure controlled terms have identifiers

### LINKED ISSUE #110: Controlled terms incorrectly are treated as literals
Subject level controlled terms (sex, subjectGroup) are currently treated as literals.

https://github.com/neurobagel/bagel-cli/blob/a9ea5054972bc6f4c7ac6652dce5762d60ed0ae6/bagel/models.py#L51-L52

This isn't good, because the "literal" values they receive are (namespaced) IRIs that refer to the controlled terms they are mapped to. What is needed is for these attributes to be given a type other than `str` or `float`, specifically a pydantic class with an `identifier` attribute. We already have that for `Image`, `Diagnosis`, and `Assessment`, e.g.:

https://github.com/neurobagel/bagel-cli/blob/a9ea5054972bc6f4c7ac6652dce5762d60ed0ae6/bagel/models.py#L35-L37

None of these classes currently have any attributes other than "identifier" and "schemaKey". So I think it would make sense to

- [x] Refactor the `Image`, `Diagnosis`, and `Assessment` classes to a more generic `"ControlledTerm"` class
- [x] refactor the `"schemaKey"` attribute of this class to be a mandatory, user defined string value. This will allow users to specify the `@type` of the instance of `ControlledTerm` they are creating
- [x] change the CLI so it invokes `ControlledTerm` for all controlled terms and also provides the correct schemaKey
- [x] Test that

surchs:
Closes #110 

As part of this PR, the `example_synthetic.jsonld` test target had to be re-generated because the data model has changed, and the old  target file therefore wouldn't pass validation. I did a quick visual check, but would be good to check this again - after all it's a bit circular.

### DIFF
--- a/bagel/tests/data/example_synthetic.jsonld
+++ b/bagel/tests/data/example_synthetic.jsonld
@@ -9,9 +9,9 @@
       "@id": "bg:hasContrastType"
     },
     "schemaKey": "@type",
-    "Assessment": "bg:Assessment",
     "Bagel": "bg:Bagel",
     "BaseModel": "bg:BaseModel",
+    "ControlledTerm": "bg:ControlledTerm",
     "Dataset": "bg:Dataset",
     "label": {
       "@id": "bg:label"
@@ -19,7 +19,6 @@
     "hasSamples": {
       "@id": "bg:hasSamples"
     },
-    "Diagnosis": "bg:Diagnosis",
     "Image": "bg:Image",
     "Session": "bg:Session",
     "filePath": {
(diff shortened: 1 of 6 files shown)


bagel/tests/test_cli_pheno.py
```diff
@@ -116,7 +116,40 @@ def test_diagnosis_and_control_status_handled(
...
+    pheno = load_test_json(tmp_path / "pheno.jsonld")
+
+    for sub in pheno["hasSamples"]:
+        if attribute in sub.keys():
+            value = sub.get(attribute)
+            if not isinstance(value, list):
+                value = [value]
+            assert all(
+                ["identifier" in entry for entry in value]
+            ), f"{attribute}: did not have an identifier for subject {sub} and value {value}"
```
alyssadai, line 152: Ou, I like the printing of the `sub` and `value`!

bagel/tests/data/example_synthetic.jsonld
```diff
@@ -9,17 +9,16 @@
       "@id": "bg:hasContrastType"
     },
     "schemaKey": "@type",
-    "Assessment": "bg:Assessment",
     "Bagel": "bg:Bagel",
     "BaseModel": "bg:BaseModel",
+    "ControlledTerm": "bg:ControlledTerm",
```
alyssadai, line 14: Do we actually need this in the context? Would "ControlledTerm" ever appear as an actual annotation?

PR #120, opened 2023-04-17
[FIX]: raise exception if subject or session ids are missing for any rows

### LINKED ISSUE #91: unhandled error: when participant_id column has missing values, bagel pheno crashes
If the input pheno.tsv file has missing values / empty strings / nan in one of the participant_id columns, things break in unexpected and unhandled ways.

surchs:
Closes #91


bagel/pheno_utils.py
```diff
@@ -237,6 +237,12 @@ def find_unused_missing_values(
     return all_unused_missing_vals
 
 
+def get_rows_with_empty_strings(df: pd.DataFrame, columns: list) -> list:
+    """Returns the ids of rows with empty strings"""
+    empty_row = df[columns].applymap(lambda cell: cell == "").apply(lambda row: any([value for value in row]), axis=1)
```
alyssadai, line 242: This is a cool way to check for 'empty' cells! One thought - we're currently using `""` as our empty indicator due to our decision not to interpret `nan`s during the `read_csv` step / keeping everything as strings, right? I've read that `apply` / `applymap` can be time costly on large dataframes - what about replacing all the "" with `nan` first and then using `.isnull()` / `.isna()` on the dataframe to get the indices?

bagel/tests/data/README.md
```diff
@@ -2,19 +2,20 @@
...
+| 9            | invalid, based on example 6 but contains an unannotated value for `group`                                                                                                | valid, based on example 6                                                            | fail               |
+| 10           | valid, same as example 6                                                                                                                                                 | valid, based on example 6 but contains extra `"MissingValues"` not found in the .tsv | pass, with warning |
+| synthetic    | valid, has `participant` and `session` IDs corresponding to the [`synthetic` example BIDS dataset](https://github.com/bids-standard/bids-examples/tree/master/synthetic) | valid                                                                                | pass               |
+| 11           | invalid, ex 6 with missing entries in `participant_id` columns                                                                                                           | valid, based on example 6                                                            | fail               |
```
alyssadai, line 19:
```suggestion
| 11           | invalid, ex 6 with missing entries in `participant_id` and `session_id` columns                                                                                                           | valid, based on example 6                                                            | fail               |
```

alyssadai (approved):
Looks great @surchs! A couple 🍒 and clarification q's below.

bagel/pheno_utils.py
surchs: Ah, that's an interesting point. I'm against applying this fix now because we haven't encountered a slowdown because of these `apply` statements, so let's see if this becomes an issue first. In addition, right now our bottleneck is almost certainly `pybids` with its sqlite initialization. But we could add a comment here with your suggestion in case things do slow down, so we remember it (or remind ourselves to check this sections as the slowdown culprit in a profile).

PR #122, opened 2023-04-19
Configure workflow to push to DockerHub + add instructions for Singularity

### LINKED ISSUE #118: Configure GitHub action to build + push Docker image to DockerHub, document Singularity image build
- [x] Set up automatic Push to https://hub.docker.com/r/neurobagel/bagelcli/tags
  - see https://docs.docker.com/language/golang/configure-ci-cd/
- [x] Describe how to build a Singularity image from the DockerHub image in README

alyssadai:
Closes #118

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,32 @@
+name: Continuous Integration
+
+on:
+  push:
+    branches:
+      - "main"
+  workflow-dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      -
+        name: Build and push
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: ./Dockerfile
+          push: true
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
surchs, line 1: The file should be called either `docker_build` or `cd` or something with "delivery" to differentiate it from a testing workflow.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,32 @@
+name: Continuous Integration
```
surchs, line 1: more fitting here would be "delivery", because the workflow is about building a runnable version of the tool (see https://www.atlassian.com/continuous-delivery/principles/continuous-integration-vs-delivery-vs-deployment). 

Because we'll see this name on the check summary page, I would be more descriptive and say something like "build and push to dockerhub" or something similar

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,32 @@
...
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      -
+        name: Build and push
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: ./Dockerfile
+          push: true
+          tags: ${{ secrets.DOCKERHUB_USERNAME }}/bagelcli:latest
```
surchs, line 32: This is cool, but it does remind me that we need to figure out how / when to start releasing (automatically?) very soon now. `latest` isn't a great tag target for deployment, we'll want to pin specific versions. I'll make an issue for us to figure this out and clarify the deployment process in response to a new docker-build / release

surchs (approved):
Nice. Very exciting @alyssadai. I'm approving, but please change the workflow name before merging to clarify this is about building. Thanks for updating the readme as well! :cook:

.github/workflows/ci.yaml
surchs: https://github.com/neurobagel/project/issues/86

PR #123, opened 2023-04-19
codespell: config and workflow

yarikoptic:
Similar to https://github.com/neurobagel/annotation_tool/pull/410 but without typos fixed -- good job on keeping it clean. Let's keep it that way!

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,19 @@
+---
+name: Codespell
+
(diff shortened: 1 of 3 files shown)


alyssadai:
Thanks again @yarikoptic for this PR! I've just made a small fix so the indentation of the `codespell` hook is consistent with the others, similar to https://github.com/neurobagel/proc_dash/pull/32.

PR #127, opened 2023-05-09
Align terms to naming standards, refactor use of namespaces

### LINKED ISSUE #89: Neurobagel mappings use inconsistent capitalization
`sex` and `diagnosis` use lowercase mapped terms, all other cateogories use capitalized mapped terms, [see here](https://github.com/neurobagel/bagel-cli/blob/7f85176c310180bd4cfd8bbf6af8597f50cffb0d/bagelbids/mappings.py#L17).

We should:

- [ ] make all terms use the same style (probably capitalized)
- [ ] go to the documentation and update the examples so they use the correct mapped term: https://www.neurobagel.org/documentation/dictionaries/

### LINKED ISSUE #124: Align class/resource names across the generated context and data model schemaKey attributes
- [x] Ensure that the @context created by the CLI matches the schemaKey attribute we use in capitalization and spelling (i.e. avoid things like @context: ... "diagnosis" and then "schemaKey": "Diagnosis" because they won't match
- [x] Ensure that all term namespaces are correct following https://github.com/neurobagel/encarta/blob/main/docs/standards.md
- [x] Data model attributes in models.py should follow appropriate naming conventions for properties
- [x] Update test examples
  - [x] replace bids sex terms with snomed URLs

### LINKED ISSUE #125: Switch controlled term namespaces to variables
We have controlled term namespace prefixes 'hard-coded' in several places in the CLI, e.g. instances of "bg" in https://github.com/neurobagel/bagel-cli/blob/main/bagel/pheno_utils.py and https://github.com/neurobagel/bagel-cli/blob/main/bagel/models.py, which currently makes updating them a bit tedious.

As an alternative, we could create a set of namedtuples to store all of our used namespaces, i.e. `Namespace = namedtuple("Namespace", ["pf", "url"])` (pf for prefix) and access the values elsewhere in the codebase using the field names instead of the actual string prefixes for example. This way, we can also reuse the named tuples to construct the context.

alyssadai:
- Controlled terms and model class attributes (properties) have been renamed according to https://github.com/neurobagel/encarta/blob/main/docs/standards.md. Tests and example files have also been updated.
- Used namespaces are now stored in `namedtuples` and prefixes accessed through the corresponding field
- New constant created to store labels and URIs of age heuristics

Closes #124, closes #89, closes #125

### DIFF
--- a/bagel/cli.py
+++ b/bagel/cli.py
@@ -77,7 +77,7 @@ def pheno(
 
         subject = models.Subject(hasLabel=str(participant))
         if "sex" in column_mapping.keys():
-            subject.sex = models.ControlledTerm(
+            subject.hasSex = models.ControlledTerm(
                 identifier=putil.get_transformed_values(
                     column_mapping["sex"], _sub_pheno, data_dictionary
                 ),
@@ -96,14 +96,14 @@ def pheno(
                     schemaKey="SubjectGroup",
                 )
             else:
-                subject.diagnosis = [
+                subject.hasDiagnosis = [
                     models.ControlledTerm(
                         identifier=_dx_val, schemaKey="Diagnosis"
                     )
                 ]
 
         if "age" in column_mapping.keys():
-            subject.age = putil.get_transformed_values(
+            subject.hasAge = putil.get_transformed_values(
                 column_mapping["age"], _sub_pheno, data_dictionary
             )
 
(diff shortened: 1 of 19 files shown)


surchs (approved):
Looks good @alyssadai. Good to merge!

surchs (approved):
Looks good @alyssadai. Good to merge!

PR #136, opened 2023-05-10
[FIX] fix typo cd.yml

Remi-Gau:
this caused the cd workflow to systematically fail

see here
https://github.com/neurobagel/bagel-cli/actions/workflows/cd.yml


alyssadai (approved):
Perfect, thanks for catching this!

PR #140, opened 2023-05-17
[FIX] Reimplement controlled term classes

### LINKED ISSUE #129: Add identifiers of used controlled terms back into `@context`
Ever since we refactored models.py classes corresponding to controlled terms with only `identifier` and `schemaKey` attributes into a general `ControlledTerm` class (see https://github.com/neurobagel/bagel-cli/issues/110), the function which generates the context for the .jsonld (see below) no longer adds specific named instances of ControlledTerm and their URIs to the context. So, the following resources (representing types) currently can appear as schemaKeys in the .jsonld but currently are not mapped to a namespace in the created context:
- "Diagnosis"
- "Assessment"
- "Sex"
- "SubjectGroup"
- "Image"

https://github.com/neurobagel/bagel-cli/blob/827179c099a725457c16ad9c8525abca5574a03a/bagel/pheno_utils.py#L25-L48

### Fix
~- [ ] add namespace prefix to schemaKeys, e.g. `"schemaKey": "Sex"` becomes `"schemaKey": "nb:Sex"`~
- [x] **add back a separate named class for each controlled term**
  - [x] Keep ControlledTerms class and just inherit from it for each to-be-recreated classes (provide their schemaKey attribute on instantiation). This way, we will only have to add another one of these classes when we add a new variable to our data model.
  - [x] update params of `test_generate_context`
  - [x] create simple test to check that all classes have URIs for their schemaKey in the `@context`
  - [x] regenerate example_synthetic.jsonld

Not currently planned:
- currently this bug is not caught by any of our tests, since we don't "expand" any example .jsonld files to view the actual generated graph model. Maybe we could try using a jsonld parser in Python (https://github.com/digitalbazaar/pyld ?) to catch some of these non-URI `@types`

### LINKED ISSUE #138: Remove `range` as an age heuristic
`range` is no longer an accepted age heuristic following https://github.com/neurobagel/documentation/pull/16. To conform to this change, the list of heuristics in `pheno_utils.py` and the corresponding test need to be updated accordingly.

alyssadai:
Closes #129, closes #138.

- Separate classes reimplemented for controlled terms, except they inherit from a common class (`ControlledTerm`)
- New test added which checks that in a CLI-outputted .jsonld, classes provided as `schemaKey` values for an attribute also have a definition in the `@context` (will help catch future occurrences of classes not defined explicitly in `models.py` appearing in the .jsonld as a namespace-less element)
- `nb:range` age heuristic removed

## Checklist

- [X] PR has an interpretable title with a prefix (`[ENH]`, `[BUG]`, `[DOC]`, `[INFRA]`, `[MAINT]`)
- [X] PR links to Github issue with mention `Closes #XXXX`
- [X] Tests pass
- [X] Code is properly formatted

For new features:
- [X] Tests have been added

For bug fixes:
  - N/A, bug fixed in this PR was not caught by original set of tests

### DIFF
--- a/bagel/tests/test_cli_pheno.py
+++ b/bagel/tests/test_cli_pheno.py
@@ -224,16 +224,20 @@ def test_controlled_term_classes_have_uri_type(
         ],
     )
 
-    pheno = load_test_json(tmp_path / "pheno.jsonld")
+    pheno = load_test_json(
+        test_data / "example_synthetic.jsonld"
+    )  # tmp_path / "pheno.jsonld"
 
     for sub in pheno["hasSamples"]:
         for key, value in sub.items():
-            if isinstance(value, (list, dict)):
-                if not isinstance(value, list):
-                    value = [value]
-                assert all(
-                    entry["schemaKey"] in pheno["@context"] for entry in value
-                ), f"Attribute {key} for subject {sub} has a schemaKey that does not have a corresponding URI in the context."
+            if not isinstance(value, (list, dict)):
+                continue
+            if isinstance(value, dict):
+                value = [value]
+            assert all(
+                entry.get("schemaKey", "no schemaKey set") in pheno["@context"]
+                for entry in value
+            ), f"Attribute {key} for subject {sub} has a schemaKey that does not have a corresponding URI in the context."
 
 
 @pytest.mark.parametrize(
--- a/bagel/tests/test_cli_pheno.py
+++ b/bagel/tests/test_cli_pheno.py
@@ -205,6 +205,37 @@ def test_controlled_terms_have_identifiers(
             ), f"{attribute}: did not have an identifier for subject {sub} and value {value}"
 
 
+def test_controlled_term_classes_have_uri_type(
+    runner, test_data, tmp_path, load_test_json
+):
+    """Tests that classes specified as schemaKeys (@type) for subject-level attributes in a .jsonld are also defined in the context."""
+    runner.invoke(
+        bagel,
+        [
+            "pheno",
+            "--pheno",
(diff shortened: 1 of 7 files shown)


bagel/tests/test_cli_pheno.py
```diff
@@ -205,6 +205,37 @@ def test_controlled_terms_have_identifiers(
...
+            "do not care name",
+        ],
+    )
+
+    pheno = load_test_json(tmp_path / "pheno.jsonld")
+
+    for sub in pheno["hasSamples"]:
+        for key, value in sub.items():
+            if isinstance(value, (list, dict)):
+                if not isinstance(value, list):
```
surchs, line 232: Might be my lack of morning coffee, but doesn't this just mean `isinstance(value, dict)` given the line above?
If so, I think clearer would be `if not isinstance(value,  (list, dict)): continue` and then `if isinstance(value, dict): value = [value]`

bagel/tests/test_cli_pheno.py
```diff
@@ -205,6 +205,37 @@ def test_controlled_terms_have_identifiers(
...
+                    value = [value]
+                assert all(
+                    entry["schemaKey"] in pheno["@context"] for entry in value
```
surchs, line 235: :cherries:: how about `entry.get("schemaKey", "no schemakey set")` here. If we encounter a case with no `schemaKey`, we'll crash the test with a KeyError. But here we'll fail the test in a somewhat more understandable way. Not sure if schemaKey existence is already asserted somewhere else, if so, probably overkill

surchs (approved):
Looks good @alyssadai. One clarity and one :cherries: comment. Take a look and then :cook:

PR #169, opened 2023-06-17
[FIX] Resolve session path for subjects with single session

### LINKED ISSUE #168: BIDS session path not resolved when subject has only one session
For subjects with at least one session, we rely on `layout.get(..., return_type="dir")` to get the path of a given session:

https://github.com/neurobagel/bagel-cli/blob/b6d15e93e9242de170aa1b319d7588d92cd23491/bagel/bids_utils.py#L72-L79

However, this fails if the subject has **exactly** one session.

To reproduce:

```python
import bids
layout = bids.BIDSLayout("bids-examples/ieeg_motorMiller2007")
layout.get(subject="cc", session="01", target="session", return_type="dir")
Out[]: []

# BUT:
layout.get_sessions(subject="cc")
Out[]: ['01']

```

This is very similar to https://github.com/bids-standard/pybids/issues/978.

We should do two things:
1.  stop using `return_type="dir"` and replace it with building the path directly (as suggested in the issue)
2. being good neighbours and going over to pybids and see if we can help fix this `return_type="dir"` issue, e.g. by deprecating it / providing docs.

surchs:
Closes #168

### DIFF
--- a/bagel/bids_utils.py
+++ b/bagel/bids_utils.py
@@ -62,20 +62,10 @@ def get_session_path(
             # switch to using layout.get() snippet below to fetch subject path.
             bids_dir
             / f"sub-{bids_sub_id}"
-            # layout.get(
-            #     subject=bids_sub_id,
-            #     target="subject",
-            #     return_type="dir",
-            # )[0]
         )
     else:
-        session_path = Path(
-            layout.get(
-                subject=bids_sub_id,
-                session=session,
-                target="session",
-                return_type="dir",
-            )[0]
+        session_path = (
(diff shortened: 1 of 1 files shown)


alyssadai (approved):
Looks good @surchs! 

I think we can now actually switch to purely using `layout.root` to fetch the bids directory path in this util function (`get_session_path`), which may simplify the function definition/call further. Opened a refactoring iss for this, https://github.com/neurobagel/bagel-cli/issues/171.

alyssadai:
Given unavailability of PR assignee this week, will merge on their behalf.

PR #185, opened 2023-08-23
[ENH] Pinning dependencies for deterministic Docker builds

### LINKED ISSUE #179: [BUG] Bagel-cli doesn't work after installation

- [ ] I have searched the existing issues

### Expected Behavior

_No response_

### Current Behavior

_No response_

### Error message
```
(NeuroBagel) maellef@Coucou:~/git/bulk_annotations$ bagel --help
Traceback (most recent call last):
  File "/home/maellef/env/NeuroBagel/bin/bagel", line 5, in <module>
    from bagel.cli import bagel
  File "/home/maellef/git/bagel-cli/bagel/cli.py", line 9, in <module>
    import bagel.bids_utils as butil
  File "/home/maellef/git/bagel-cli/bagel/bids_utils.py", line 6, in <module>
    from bagel import mappings, models
  File "/home/maellef/git/bagel-cli/bagel/models.py", line 12, in <module>
    class Bagel(BaseModel, extra=Extra.forbid):
  File "/home/maellef/git/bagel-cli/bagel/models.py", line 16, in Bagel
    identifier: str = Field(
  File "/home/maellef/env/NeuroBagel/lib/python3.10/site-packages/pydantic/fields.py", line 755, in Field
    raise PydanticUserError('`regex` is removed. use `pattern` instead', code='removed-kwargs')
pydantic.errors.PydanticUserError: `regex` is removed. use `pattern` instead

For further information visit https://errors.pydantic.dev/2.0.3/u/removed-kwargs
```
### Environment

- OS: Ubuntu 22.04.2 lts
- Python/Node version: 3.10.6

### How to reproduce

try installing without container

### Anything else?

_No response_

### LINKED ISSUE #184: Add requirements.txt pinning all versions of dependencies
Ensures version bumps don't affect our codebase behaviour in unexpected ways.
See also: #180 and #179

- [x] Create pinned dependencies for Docker
- [ ] Prevent Dependabot from bumping Pydantic (to 2)
- [x] Protect the main branch from direct commits 👀 
- [x] Update Dockerfile to use new requirements.txt

surchs:
Closes #184
Closes #179

Changes proposed:

- pin pydantic to < major version 2
- add `pip-tools` to dev dependencies because we now create our python lockfile with `pip-compile` from the `pip-tools` package.
- use `pip-compile` to add a `requirements.txt` lockfile that pins all of our dependencies
- update the Dockerfile to install dependencies from the lockfile
- update README to explain how to update the lockfile

## Things left to do after this PR:

~~I didn't update the setup instructions for local development. That is because `pip-compile` makes `requirements.txt` for the main dependencies defined in `setup.cfg` but not for the dev-dependencies. There might be a way to get it to do both, but I haven't figured that out yet.~~

~~The result is that when you follow our instructions for setting up a local dev env, you will end up with something different than what we are building the Docker image with (and thus running in prod). I think our options are:~~

- ~~make a big flag to point this out~~
- ~~open an issue to fix this later~~
- ~~find a solution now and merge it with this PR~~

edit: all addressed by @alyssadai in https://github.com/neurobagel/bagel-cli/pull/185#issuecomment-1690459566


alyssadai:
@surchs re: your edit, the `--extra` flag of [`pip-compile`](https://pip-tools.readthedocs.io/en/latest/cli/pip-compile/#pip-compile) looks like a way to create another `requirements.txt` with extra dev dependencies 🕵️:
```
--extra TEXT                    Name of an extras_require group to install;
                                        may be used more than once
```

surchs:
Nice find @alyssadai, agree that's the right way to do it!

PR #212, opened 2023-09-22
[FIX] Gracefully handle user providing phenotypic file as .csv

### LINKED ISSUE #90: If user provides .csv instead of .tsv, bagel pheno CLI fails with nice looking but incorrect error
If I provide `bagel pheno` with a .csv instead of a .tsv file (i.e. comma separated data instead of tab separated), it crashes with 

```
LookupError: The provided data dictionary and phenotypic file are individually valid, but are not compatible. Make sure that you selected the correct data dictionary for your phenotyic file. Every column described in the data dictionary has to have a corresponding column with the 
same name in the phenotypic file
```

That's only a little bit better than crashing somewhere completely unexpected in the code, and definitely worse than crashing when the .tsv is loaded with some informative error message. The reason for it not crashing sooner is that a .csv is technically a valid .tsv that has only one column / no tabs anywhere.

I think we should:

- [x] validate that the pheno file has a .tsv file ending (not a guarantee, but a good start). we should probably fire a warning
- [x] if the loaded pheno file has only one column, then we can probably error out. Because the best case is that it only contains the participantID column. But a more likely case is that something went wrong with parsing the pheno file

surchs:
Closes #90 

- raise errors if
  - user provides a file ending in `.csv`
  - user provides a `.tsv` file, that really is just a repackaged `.csv` file -> we're not fooled :eyes: 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


bagel/pheno_utils.py
```diff
@@ -35,6 +39,30 @@ def validate_portal_uri(portal: str) -> Optional[str]:
...
+            input_p, sep="\t", keep_default_na=False, dtype=str
+        )
+
+        if pheno_df.shape[1] > 1:
+            # We only extracted one column, this might mean that we have a sneaky .csv
+            return pheno_df
+        elif len(pheno_df.columns[0].split(",")) == 1:
+            warnings.warn(
+                f"Your phenotypic input file {input_p} has only one column."
+                " Although this is not invalid, such a file is not valid."
```
alyssadai, line 55: > " Although this is not invalid, such a file is not valid."

I think I get what you're trying to say, but this part of the warning is a bit confusing as it seems contradictory.

How about keeping it simple:
```suggestion
                " Such a file is not a valid Neurobagel phenotypic data file."
```

bagel/tests/test_cli_pheno.py
```diff
@@ -255,6 +255,41 @@ def test_unused_missing_values_raises_warning(
...
+    ],
+)
+def test_providing_csv_file_raises_error(
+    pheno_file,
+    dictionary_file,
+    runner,
+    test_data,
+    tmp_path,
+):
+    """Providing a .csv file or a file with .tsv extension but incorrect encoding should be handled with an infomrative error."""
```
alyssadai, line 272: typo missed by codespell:

```suggestion
    """Providing a .csv file or a file with .tsv extension but incorrect encoding should be handled with an informative error."""
```

bagel/tests/data/README.md
```diff
@@ -2,24 +2,25 @@
...
+| 12           | Valid, same as example 2                                                        | Valid, based on example 2 but missing BIDS "Levels" attribute for group column       | Pass, with warning |
+| 13           | Valid, same as example_synthetic                                                | Valid, based on example_synthetic but with mismatched levels for group column        | Pass, with warning |
+| 14           | Valid, same as example 2                                                        | Valid, based on example 2, but with an extra column annotation without Neurobagel    | Pass               |
+| 15           | Valid, same as example 2                                                        | Invalid, based on example 2, but participant ID column lacks Neurobagel annotations  | Fail               |
+| 16           | Invalid, same as example2.csv, but with a sneaky .tsv file ending               | Valid, same as example2                                                              | fail               |
```
alyssadai, line 23: Just a thought -

With larger README markdown tables like this I think we're starting to see the downsides of using an auto-table formatter in our code editor. I realized the reason for the messy diffs is because generally the entire table is adjusted when a new row is added (assuming it has contents of a different length than the existing rows).

I think in future, except for very small tables it may be better to just turn off auto-formatting and stick to a basic markdown table format, e.g.

```markdown
column 1 | column 2 | column 3
---- | ---- | ----
my | first | row
my | second longer | row
my | third | row
```
This will restrict diffs to just the modified/added rows. It won't look as neat in markdown, but even with auto-formatting, as the table gets bigger, the reformatted table still ends up with messy text wrapping except on a very wide screen anyways.

alyssadai (commented):
Thanks @surchs for addressing this!

Everything seems good functionally, but I've left a few questions and 🍒 comments below I think would be worth having another look at.

bagel/pheno_utils.py
```diff
@@ -35,6 +39,30 @@ def validate_portal_uri(portal: str) -> Optional[str]:
...
 
+def load_pheno(input_p: Path) -> pd.DataFrame | None:
+    """Load a .tsv pheno file and do some basic validation."""
+    if input_p.suffix != ".csv":
+        pheno_df: DataFrame = pd.read_csv(
+            input_p, sep="\t", keep_default_na=False, dtype=str
+        )
+
+        if pheno_df.shape[1] > 1:
+            # We only extracted one column, this might mean that we have a sneaky .csv
```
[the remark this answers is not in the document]
surchs: Yes, thanks for catching this. I think I missed this during a refactor

bagel/pheno_utils.py
```diff
@@ -35,6 +39,30 @@ def validate_portal_uri(portal: str) -> Optional[str]:
     return portal
 
 
+def load_pheno(input_p: Path) -> pd.DataFrame | None:
+    """Load a .tsv pheno file and do some basic validation."""
+    if input_p.suffix != ".csv":
```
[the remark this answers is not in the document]
surchs: Hhm. Yeah, I think that's a good idea!

bagel/pheno_utils.py
```diff
@@ -35,6 +39,30 @@ def validate_portal_uri(portal: str) -> Optional[str]:
...
+        pheno_df: DataFrame = pd.read_csv(
+            input_p, sep="\t", keep_default_na=False, dtype=str
+        )
+
+        if pheno_df.shape[1] > 1:
+            # We only extracted one column, this might mean that we have a sneaky .csv
+            return pheno_df
+        elif len(pheno_df.columns[0].split(",")) == 1:
```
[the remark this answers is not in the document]
surchs: Good point, I agree this is confusing. Let me take another look with the broader file type filtering (i.e. `"Only .tsv"` instead of `"not .csv" and "not .txt"`). Maybe that'll help

bagel/tests/data/README.md
surchs: Agree. These examples are also getting a bit long.

bagel/pheno_utils.py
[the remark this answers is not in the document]
surchs: Having looked at this again, I think the implementation was just wrong here. What I wanted to see was:

- do you pretend to be a tsv
- and have only one column
- but if I split your header by comma, I get more than one element

-> congrats, you might suffer from undeclared .csv-ness. 
My check was something different

bagel/pheno_utils.py
```diff
@@ -35,6 +39,39 @@ def validate_portal_uri(portal: str) -> Optional[str]:
...
+                f"you have accidentally renamed a .csv file as a .tsv"
+                " Please provide a valid .tsv pheno file!"
+            )
+        else:
+            warnings.warn(
+                f"Your phenotypic input file {input_p} has only one column."
+                " Such a file is not a valid Neurobagel phenotypic data file."
+                " Please check your input again!"
+            )
+            return pheno_df
```
alyssadai, line 67: 🍒 Just curious - any reason you chose to raise a warning here instead of just erroring out and exiting the program?

I don't mind letting single-column TSVs still go through the CLI if there is nothing else apparently wrong with the contents, but if this is the intended behaviour we may want to change the warning phrasing to be different than the actually _invalid_ input scenario above (which _does_ error out), e.g.: 

```suggestion
                f"Your phenotypic input file {input_p} has only one column."
                " We recommend double checking your input."
            )
            return pheno_df
```

alyssadai (approved):
Two 🍒 suggestions, otherwise looks good to go!!

bagel/pheno_utils.py
surchs: No I think you're right. Let's error out. 

Speaking of that, I noticed you used this syntax recently for erroring: 

https://github.com/neurobagel/bagel-cli/blob/4434745de050c75157635d81fe877e5c3531b471/bagel/utility.py#L16-L21

Do you think we should adopt that across the board? What's the benefit?

alyssadai: @surchs aha, I adopted that syntax because I was looking at this docs [page](https://typer.tiangolo.com/tutorial/terminating/) and thought our scenario of "file already exists" was similar to the example they provided and so wanted to try it. 😛

I think the benefit is that the output message looks really clean, and there's no traceback or anything like that which makes sense for an upfront sanity check like this. On the other hand, I think this kind of termination is probably not the most suitable for handling more complex input errors - the default exit code is still 0, so it's not treated like other exceptions like `ValueError` etc unless we manually pass in error codes. This might make testing a little trickier as well: from my brief playing around with it for this PR, our normal syntax of using `with pytest.raises` doesn't seem to capture these exit exceptions (?).

All in all, I think most of the custom errors (TSV vs JSON incompatibility, missing keys, etc.) we catch probably still are conceptually better handled as normal exceptions, but we can revisit this if needed as part of https://github.com/neurobagel/bagel-cli/issues/218.

surchs: Let's leave this for #218

PR #216, opened 2023-09-25
[ENH] Provide overwritable output file path instead of directory path arg

### LINKED ISSUE #157: Make output .jsonld file name user configurable
currently our output .jsonld files all get the same name:

- pheno: pheno.jsonld
- bids: pheno_bids.jsonld

That's not super informative, and it doesn't work well if we run the CLI for more than one dataset and put all outputs in the same place. So we need:

- [x] keep the current defaults in order not to break things
- [x] replace the output directory option with a file-path that includes the name (but keep the name `--output`)
- [x] make sure the file does not exist yet, but if it does, only allow if `--overwrite` is also true
- [x] add a `--overwrite` flag to overwrite
- add test that:
  - [x] if CLI not invoked with value for `--output`, an output file with the default name is created in the current dir
  - [x] if CLI not invoked with `--overwrite` but output file already exists, a warning is emitted

alyssadai:
Closes #157

- `--output` argument changed to specify a file path instead of a directory, and is no longer required for either command (defaults if unset are `pheno.jsonld`/`pheno_bids.jsonld`)
- `--overwrite` flag added which will overwrite an existing output file if set
  - if unset and the output file exists, program exits with informative message 
- new fixtures created for default output file path to use in tests

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/bagel/cli.py
+++ b/bagel/cli.py
@@ -9,7 +9,7 @@ from pydantic import ValidationError
 import bagel.bids_utils as butil
 import bagel.pheno_utils as putil
 from bagel import mappings, models
-from bagel.utility import load_json
+from bagel.utility import check_overwrite, load_json
 
 bagel = typer.Typer()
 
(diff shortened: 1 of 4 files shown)


surchs (approved):
Very nice! :cook:

PR #219, opened 2023-10-03
[MNT] Capitalized age heuristics

### LINKED ISSUE #128: Capitalize age heuristics
Age heuristics are currently lowercase. According to the new naming standards, these probably are equivalent to resources and thus should be capitalized, e.g. nb:Float instead of nb:float (?)

_Originally posted by @surchs in https://github.com/neurobagel/bagel-cli/pull/127#discussion_r1188951514_

TODO:

- [x] Find names that better reflect that these are transformations. @alyssadai suggested `nb:FromISO8061` for example
*`ISO8061` implies that's the format the age gets stored as (target of the transformation) however, `FromISO8061` makes it explicit that it's a transformation from `ISO8061` to some other format*
- [x] Turn all of the transformations into PascalCase
- [x] Update 
  - [x] all of the corresponding docs
  - [x] CLI logic for transformation and corresponding tests
- [x] create corresponding issue on the annotation tool to address as part of reviving that

rmanaem:
Closes #128 

- Capitalized age heuristics
- Update CLI logic and related tests
- Created issues in [annotation_tool](https://github.com/neurobagel/annotation_tool/issues/561), [documentation](https://github.com/neurobagel/documentation/issues/101), and [neurobagel_example](https://github.com/neurobagel/neurobagel_examples/issues/12) repos to update the heuristics accordingly

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/bagel/pheno_utils.py
+++ b/bagel/pheno_utils.py
@@ -18,11 +18,11 @@ from bagel.mappings import COGATLAS, NB, NCIT, NIDM, SNOMED
 DICTIONARY_SCHEMA = dictionary_models.DataDictionary.schema()
 
 AGE_HEURISTICS = {
-    "float": NB.pf + ":float",
-    "int": NB.pf + ":int",
-    "euro": NB.pf + ":euro",
-    "bounded": NB.pf + ":bounded",
-    "iso8601": NB.pf + ":iso8601",
+    "FromFloat": NB.pf + ":FromFloat",
+    "FromInt": NB.pf + ":FromInt",
+    "FromEuro": NB.pf + ":FromEuro",
+    "FromBounded": NB.pf + ":FromBounded",
+    "FromISO8601": NB.pf + ":FromISO8601",
 }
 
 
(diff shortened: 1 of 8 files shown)


bagel/pheno_utils.py
```diff
@@ -18,11 +18,11 @@
 DICTIONARY_SCHEMA = dictionary_models.DataDictionary.schema()
 
 AGE_HEURISTICS = {
-    "float": NB.pf + ":float",
-    "int": NB.pf + ":int",
-    "euro": NB.pf + ":euro",
-    "bounded": NB.pf + ":bounded",
-    "iso8601": NB.pf + ":iso8601",
+    "FromFloat": NB.pf + ":FromFloat",
```
alyssadai, line 21: Unless it breaks something, I suggest we keep the original simple dict keys `"float"`, `"int"`, etc as these are just meant to be internal aliases for the URIs and never enter the actual graph data files. This will make our lives easier down the line since they're shorter to type + means we don't need to update the functions that use this dict when/if we update the actual URI values.

alyssadai (approved):
One 🍒, otherwise good to go!

PR #227, opened 2023-10-21
[REF] Refactor phenotypic/BIDS subjects check and add more intermediate stdout for `bagel bids`

### LINKED ISSUE #190: Check that all BIDS subjects are in pheno.jsonld prior to BIDS parsing
The goal is to catch + return errors to do with non-matching BIDS and pheno subject lists _before_ slow pybids processing.

Some notes
- This is actually not easily done using pybids without instantiating a `BIDSLayout` object (which is the slowest step anyways) as the layout is needed for subject indexing (e.g., `.get_subjects()`)
- Some things that _do_ speed up the `BIDSLayout` step:
  - passing in a `BIDSLayoutIndexer(index_metadata=False)` ✅ (See: https://bids-standard.github.io/pybids/generated/bids.layout.index.BIDSLayoutIndexer.html#bids.layout.index.BIDSLayoutIndexer
  - BUT, for overall speed, we probably don't want to have to call `BIDSLayout` twice (once with `index_metadata=False` and once without)
- Using the `ignore` param with a regex pattern to match paths except top-level directories starting with `sub-` doesn't seem to work - the resulting layout always has 0 subjects
  - probably because pybids indexes subjects based on _file_ metadata

### Steps to implement:

- [x] use a simple directory traversal to match `sub-` directory names
- [x] output more intermediate messages to user during slow BIDS step such as
  - when initial check of subjects / validation of pheno JSONLD has successfully completed
  - before and after slow BIDS dataset initialization (e.g., "reading BIDS dataset")
  - when the Neurobagel merging process for the parsed BIDS metadata and phenotypic annotations begins

### Related issues
- [ ] update CLI docs with messages to look out for when running BIDS command

alyssadai:
Closes #190 

- Check for BIDS subject IDs missing from the phenotypic JSONLD _before_ slow `BIDSLayout` initialization
  - This is done by simply extracting BIDS subdirectory names (see #190 issue description for why pybids functions are limited here)
- Print more intermediate messages during `bagel bids` command to indicate program progress

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/bagel/bids_utils.py
+++ b/bagel/bids_utils.py
@@ -11,6 +11,19 @@ def map_term_to_namespace(term: str, namespace: dict) -> str:
     return namespace.get(term, False)
 
 
+def get_bids_subjects_simple(bids_dir: Path) -> list:
+    """Returns list of subject IDs (in format of sub-<SUBJECT>) for a BIDS directory inferred from the names of non-empty subdirectories."""
+    bids_subject_list = []
+    for path in bids_dir.iterdir():
+        if (
+            path.name.startswith("sub-")
+            and path.is_dir()
+            and any(path.iterdir())
+        ):
+            bids_subject_list.append(path.name)
+    return bids_subject_list
+
+
 def check_unique_bids_subjects(pheno_subjects: list, bids_subjects: list):
     """Raises informative error if subject IDs exist that are found only in the BIDS directory."""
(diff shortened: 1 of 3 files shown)


rmanaem (approved):
:cook:

Issue #183, opened 2023-08-22
Build Docker on release, not (just) push


surchs:
Currently we build and push a docker image every time we push to `main`: https://github.com/neurobagel/bagel-cli/blob/1d4cb5217813cedfbe6d78fcb2821fac72059089/.github/workflows/cd.yml

That only really makes sense for testing builds, not for things we deploy to production (c.f. https://github.com/neurobagel/documentation/wiki/Release-process#what-happens-next-todo).

What we should do is:

- [ ] make a second deploy workflow that listens for "release: released" (see [here](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=released#release)), have that build and tag with release number
- [ ] only deploy those explicitly tagged images to production
- [ ] release often (so we have nice things running in production), look into https://github.com/containrrr/watchtower or https://www.ansible.com/ (?)

surchs:
This now works: https://github.com/neurobagel/bagel-cli/blob/main/.github/workflows/build_docker_on_release.yml

PR #252, opened 2023-12-13
[CI] Build docker image on release

surchs:

- Related to #183

So far, we only made releases with auto. However because the release is done by a bot, this will not trigger our "release-published" hook that we rely on to build and push our docker images. So this PR

- builds and pushes the docker image directly inside of the release workflow (could also be a separate workflow that is just called from there)
- gets the release version and tags the docker image with it
- removes the old (on release) docker build workflow

## Checklist

- [ ] PR has an interpretable title with a prefix (`[ENH]`, `[FIX]`, `[REF]`, `[TST]`, `[CI]`, `[MNT]`, `[INF]`, `[MODEL]`, `[DOC]`) _(see https://neurobagel.org/contributing/pull_requests for more info)_
- [ ] PR links to GitHub issue with mention `Closes #XXXX`
- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


surchs:
Closing in favour of the steps outlined here: https://github.com/neurobagel/planning/issues/64

PR #313, opened 2024-07-17
[FIX] Handle dashes (/any chars that are not python variable valid) in column names

### LINKED ISSUE #312: [BUG] Dash in a column name breaks the pheno parsing

- [X] I have searched the existing issues

### Expected Behavior

When I have a column `BIDS-ID` in my `.json` data dictionary

### Current Behavior

The CLI breaks at the pheno command

### Error message

```
UndefinedVariableError: name 'BIDS' is not defined
```
Likely because the column name is split at the `-` somehow

### Environment

_No response_

### How to reproduce

Make a column name with `BIDS-ID` in the `.json` data dictionary (and the .tsv)

### Anything else?

It looks like when accessing (?) the data dictionary key, for some reason only the first part before the hyphen is read. As a result, the column "BIDS" is searched for in the TSV, which returns no matches. This hyphen weirdness might be due to the utilities we're using to parse JSON in the CLI, but needs some more investigating.

Switching "BIDS-ID" to "BIDS_ID" in both the TSV & JSON resolved the error.

alyssadai:
- Closes #312

- when using [pd.DataFrame.query](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.query.html), column names are now quoted w/ backticks so that characters not valid in Python variables can be read
  - prevents column names from being truncated during the query
- README table of example data also reformatted for readability/easier updating

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:
- [x] There is at least one test that would fail under the original bug conditions.

### DIFF
--- a/bagel/cli.py
+++ b/bagel/cli.py
@@ -97,7 +97,9 @@ def pheno(
     session_column = column_mapping.get("session")
 
     for participant in pheno_df[participants].unique():
-        _sub_pheno = pheno_df.query(f"{participants} == '{str(participant)}'")
+        _sub_pheno = pheno_df.query(
+            f"`{participants}` == '{str(participant)}'"
+        )
 
         sessions = []
         for session_row_idx, session_row in _sub_pheno.iterrows():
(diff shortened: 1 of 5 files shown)


rmanaem (approved):
:cook:

PR #329, opened 2024-07-19
[CI] Release the CLI

alyssadai:
## Release Notes

This release introduces short option names for CLI commands and fixes a bug that prevented hyphens in column names from being parsed.


alyssadai:
NOTE: This was a test PR to trigger a release, which was then reverted. Adding the `skip-release` label to avoid this PR contaminating future releases.

PR #330, opened 2024-07-19
[CI] Release the CLI

alyssadai:
## Release Notes

This release introduces short option names for CLI commands and fixes a bug that prevented hyphens in column names from being parsed.


alyssadai:
**NOTE:** This PR is meant to trigger a release, and includes a change to use an access token during the Checkout step, as suggested in: https://stackoverflow.com/questions/69263843/how-to-push-to-protected-main-branches-in-a-github-action

Issue #218, opened 2023-09-26
Shorten tracebacks for errors


alyssadai:
The tracebacks are currently quite long and difficult to parse. We should try and limit the error output to the informative message and what's necessary for the user to identify the issue in their input.

We may want to also consider https://typer.tiangolo.com/tutorial/terminating/#exit-with-an-error where it makes sense for upfront checks.

See also https://github.com/neurobagel/bagel-cli/pull/212#discussion_r1337604637

- [x] maybe make this configurable via a flag

surchs:
#162 seems related, let's handle those together and not deal with the present issue now

surchs:
it would make sense to combine this with using a proper logger for the messages. We could then also just decide to not pass the traceback to the logger when we handle the exception - I believe that would already skip the long context rich prints as part of the traceback (but not sure). Either way: logger = good

PR #337, opened 2024-08-15
[ENH] More user-friendly handling of input decoding/reading errors

### LINKED ISSUE #167: Make `ValueError` raised when jsonschema validation fails more informative
For example, if someone neglects to annotate the transformation for an age column, the CLI complains about the JSON not being a "valid Neurobagel data dictionary" but is not very specific.

Relevant section of CLI code:

https://github.com/neurobagel/bagel-cli/blob/13d92942cd939733c2f4dc9798a57ba5e1a78e51/bagel/pheno_utils.py#L351-L358

We should aim to highlight the problematic section of the JSON for the user, so they do not have to try and parse a long traceback.

### LINKED ISSUE #314: [BUG] TSV file with non-utf-8 encoding produces cryptic UnicodeDecodeError

- [X] I have searched the existing issues

### Expected Behavior

When running the `bagel pheno` command on a TSV and JSON that were successfully processed by the annotation tool, a JSONLD is successfully generated.

### Current Behavior

CLI exits with a UnicodeDecodeError and traceback, but it is not immediately clear from the error that it is due to a file encoding problem (or which input file is the culprit).

### Error message

`UnicodeDecodeError: utf-8 can't decode byte 0xeb in position ...: invalid continuation byte`

### Environment

- OS: Linux
- Python/Node version:

### How to reproduce

- Run `bagel pheno` on a TSV with a non-utf encoding (e.g., `latin1`) and the corresponding JSON from the annotation tool

### Anything else?

For simplicity and to avoid slowing down the CLI, we should probably resolve this for now by:
- [x] explicitly specifying the utf-8 encoding when reading in the TSV (and JSON?)
- [x] catch and emit an informative error to the user when either input is not in UTF-8
- [x] provide a link or name of a tool the user can use to convert their files

To avoid the TSV / JSON potentially becoming incompatible after encoding conversion, we should implement a similar check in the annotation tool.

alyssadai:
- Closes #314 
- Closes #167 

- Explicitly require UTF-8 encoding of inputs for the `bagel pheno` command
- Print informative error message (without a traceback!) when a UnicodeDecodeError or JSONDecodeError are encountered
- Return more details about offending entry in data dict when jsonschema validation fails

**For reviewer: Any thoughts on whether this should have the `pr-patch` or `pr-bug-fix` label?**

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/bagel/pheno_utils.py
+++ b/bagel/pheno_utils.py
@@ -366,9 +366,21 @@ def validate_data_dict(data_dict: dict) -> None:
     try:
         jsonschema.validate(data_dict, DICTIONARY_SCHEMA)
     except jsonschema.ValidationError as e:
+        # TODO: When *every* item in an input JSON is not schema valid,
+        # jsonschema.validate will raise a ValidationError for only *one* item among them.
+        # Weirdly, the item that is chosen can vary, possibly due to jsonschema internally processing/validating
+        # items in an inconsistent order.
+        # You can reproduce this by running `bagel pheno` on the example_invalid input pair.
+        # This isn't necessarily a problem as the error will not be wrong, but if we care about
+        # returning ALL invalid items, we may want to use something like a Draft7Validator instance instead.
+        # NOTE: If the validation error occurs at the root level (i.e., the entire JSON object fails),
+        # e.path may be empty. We have a backup descriptor "Entire document" for the offending item in this case.
         raise ValueError(
-            "The provided data dictionary is not a valid Neurobagel data dictionary. "
-            "Make sure that each annotated column contains an 'Annotations' key."
+            "The provided data dictionary is not a valid Neurobagel data dictionary.\n"
+            "Details:\n"
+            f"Entry that failed validation: {e.path[-1] if e.path else 'Entire document'}\n"
+            f"{e.message}\n"
+            "Tip: Make sure that each annotated column contains an 'Annotations' key."
         ) from e
 
     if get_annotated_columns(data_dict) == []:
--- a/bagel/pheno_utils.py
+++ b/bagel/pheno_utils.py
@@ -9,10 +9,12 @@ import isodate
 import jsonschema
 import pandas as pd
 import pydantic
+import typer
 from pandas import DataFrame
 from typer import BadParameter
 
 from bagel import dictionary_models, mappings, models
+from bagel import utility as utils
 from bagel.mappings import COGATLAS, NB, NCIT, NIDM, SNOMED
 
 DICTIONARY_SCHEMA = dictionary_models.DataDictionary.schema()
@@ -42,9 +44,21 @@ def validate_portal_uri(portal: str) -> Optional[str]:
 def load_pheno(input_p: Path) -> pd.DataFrame | None:
     """Load a .tsv pheno file and do some basic validation."""
     if input_p.suffix == ".tsv":
-        pheno_df: DataFrame = pd.read_csv(
-            input_p, sep="\t", keep_default_na=False, dtype=str
-        )
+        try:
+            pheno_df: DataFrame = pd.read_csv(
+                input_p,
+                sep="\t",
+                keep_default_na=False,
+                dtype=str,
+                encoding="utf-8",
+            )
+        except UnicodeDecodeError as e:
+            # TODO: Refactor once https://github.com/neurobagel/bagel-cli/issues/218 is addressed
+            typer.echo(
+                utils.file_encoding_error_message(input_p),
+                err=True,
+            )
+            raise typer.Exit(code=1) from e
 
         if pheno_df.shape[1] > 1:
             return pheno_df
--- a/bagel/pheno_utils.py
+++ b/bagel/pheno_utils.py
@@ -14,8 +14,8 @@ from pandas import DataFrame
 from typer import BadParameter
 
 from bagel import dictionary_models, mappings, models
-from bagel import utility as utils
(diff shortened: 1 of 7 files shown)


bagel/pheno_utils.py
```diff
@@ -42,9 +44,21 @@ def validate_portal_uri(portal: str) -> Optional[str]:
...
+                dtype=str,
+                encoding="utf-8",
+            )
+        except UnicodeDecodeError as e:
+            # TODO: Refactor once https://github.com/neurobagel/bagel-cli/issues/218 is addressed
+            typer.echo(
+                file_encoding_error_message(input_p),
+                err=True,
+            )
+            raise typer.Exit(code=1) from e
```
surchs, line 61: Very cool. Just out of curiosity: [rich](https://rich.readthedocs.io/en/stable/index.html) (which does the coloring of the terminal for typer afaik) does not have a way to color / make nice custom error classes? I am mainly asking because from reading (not trying) this I assume our pretty colorful text would be immediately followed by the stacktrace from hell that typer already generates, right?

bagel/pheno_utils.py
```diff
@@ -352,9 +366,21 @@ def validate_data_dict(data_dict: dict) -> None:
     try:
         jsonschema.validate(data_dict, DICTIONARY_SCHEMA)
     except jsonschema.ValidationError as e:
+        # TODO: When *every* item in an input JSON is not schema valid,
+        # jsonschema.validate will raise a ValidationError for only *one* item among them.
+        # Weirdly, the item that is chosen can vary, possibly due to jsonschema internally processing/validating
+        # items in an inconsistent order.
+        # You can reproduce this by running `bagel pheno` on the example_invalid input pair.
+        # This isn't necessarily a problem as the error will not be wrong, but if we care about
+        # returning ALL invalid items, we may want to use something like a Draft7Validator instance instead.
```
surchs, line 375: Very interesting. I wasn't aware of the difference between `jsonschema.validate` and `jsonschema.Validator.validate`. would you mind opening an issue for potentially refactoring to using the second form in the future?

bagel/pheno_utils.py
```diff
@@ -352,9 +366,21 @@ def validate_data_dict(data_dict: dict) -> None:
     try:
         jsonschema.validate(data_dict, DICTIONARY_SCHEMA)
     except jsonschema.ValidationError as e:
+        # TODO: When *every* item in an input JSON is not schema valid,
+        # jsonschema.validate will raise a ValidationError for only *one* item among them.
```
surchs, line 370: great find btw

bagel/tests/test_utility.py
```diff
@@ -60,29 +62,68 @@ def test_all_used_namespaces_have_urls(
...
-        {
-            "participant_id": {
-                "Description": "A participant ID",
-                "Annotations": {
-                    "IsAbout": {
-                        "TermURL": "nb:ParticipantID",
-                        "Label": "Unique participant identifier",
-                    }
+@pytest.mark.parametrize(
+    "partial_data_dict, invalid_item",
```
surchs, line 66: can we rename `invalid_item` to something more descriptive? Maybe `invalid_column_name`?

bagel/utility.py
```diff
@@ -4,10 +4,38 @@
...
+            err=True,
+        )
+        raise typer.Exit(code=1) from e
+    except json.JSONDecodeError as e:
+        typer.echo(
+            typer.style(
+                f"The provided data dictionary {input_p} is not valid JSON. "
+                "Please provide a valid JSON file.",
+                fg=typer.colors.RED,
+            ),
```
surchs, line 35: why can't we use  file_encoding_error_message here?

surchs (approved):
Very cool PR, thanks @alyssadai! 

I didn't try it out locally, I'm assuming this PR does not change the long stacktraces that typer is printing to the terminal on an error. I left a comment inline about whether our nicely formatted error messages would be hidden by the long stacktraces.

Also some :cherries:, take a look.

:cook:

bagel/pheno_utils.py
alyssadai: Actually, using typer.echo + typer.Exit like this hides the stacktrace entirely :) Only our pretty message is shown in this case.

So as part of https://github.com/neurobagel/bagel-cli/issues/218, we can refactor other error messages using this approach, to hide the stacktrace when it is not necessary for input validation. In comparison, elsewhere in the code where we're currently using a simple `raise` in an Except block to output a custom message, the stacktrace is shown and our custom error message appears at the very end.

And yes, I believe you can achieve the same custom message styling using `rich` [print](https://typer.tiangolo.com/tutorial/printing/#use-rich-print) instead of `typer.echo` - I just used the simpler typer.echo to start!

bagel/pheno_utils.py
alyssadai: Done! See https://github.com/neurobagel/bagel-cli/issues/338

bagel/utility.py
alyssadai: `file_encoding_error_message` specifically provides a template for the error message that is returned when a non-UTF8 encoding is detected. Since the same error message format is used for both a non-UTF8 JSON and TSV (two separate validator functions), the convenience is used function to minimize code duplication.

The above error message is only emitted in one place in the code, which is when there is a JSONDecodeError for the input data dictionary. So it wouldn't make much sense to refactor that out into a separate function, since it would pretty much just be the same as `typer.style` itself.

Issue #346, opened 2024-09-24
update `bagel bids` subcommand to play nice with new derivatives


surchs:
TODO

update the `bagel bids` command to
- [x] accept either a `pheno.jsonld` (output of `bagel pheno`) or a `pheno-derivatives.jsonld` (output of `bagel derivatives` #345) as input
- [x] update help text
- [x] find and match existing imaging sessions between a provided `pheno-derivatives.jsonld` and a raw BIDS dataset so that we do not recreate imaging sessions if the derivatives subcommand has already created them
  - [x] if session is missing, add to potentially existing "ses-nb01"?

Related to #345

alyssadai:
Closed by #349

PR #349, opened 2024-09-27
[ENH] Add derivatives command and `pipeline-catalog` submodule

### LINKED ISSUE #344: Investigate schema and examples for the Nipoppy imaging bagel
TODO

- [x] Figure out if Nipoppy already has a schema for the imaging bagel
- [x] Figure out what the near future imaging bagel schema is going to look like
- [x] Contribute a schema for the imaging bagel to nipoppy if wanted, then integrate / link this schema here (CLI repo). If nipoppy doesn't want a schema contributed, just create the schema here via pydantic
- [x] Find or create an example instance of a nipoppy imaging bagel and make it accessible here (i.e. by adding it to the neurobagel examples repo and updating our submodule of the neurobagel examples repo)
- [x] Add new prefix namespace for Nipoppy pipeline terms (allowed names & versions)

NOTES:
- Nipoppy schema is in Pydantic v2, so we should probably see if we can use the jsonschema from their model instead for compatibility w/ our v1 CLI

### LINKED ISSUE #345: Add new sub-command for bagel-CLI derivatives:
- Inputs:
    - [x] imaging bagel TSV from Nipoppy or TSV with same columns
      - [x] add check that subject ID column has no missing values
    - [x] JSONLD from bagel pheno command, OR pheno + bids command
       - [ ] Do we need a sanity check here that the file does not already contain derivative info?
    - [x] output file path
    - [x] override flag
  - Command should
    - [ ] Validate input columns from the bagel file against the Nipoppy bagel schema
    - [x] Check that pipeline names and versions are part of allowed terms
    - [x] Check that all subject IDs from the derivatives TSV are represented in the existing JSONLD, otherwise error
    - [x] If there are imaging sessions in the JSONLD, try to match them to sessions from the derivatives TSV
    - [x] If there are new sessions in the derivatives TSV, create new imaging sessions
    - [x] If a session ID is missing, create a new imaging session with a custom session label
    - [x] Extract pipeline names for a given subject-session entry in the bagel TSV, and also the completion status

alyssadai:
- Closes #344
- Closes #345 
- See also https://github.com/neurobagel/neurobagel_examples/pull/34
  - **should be merged first, currently causing test failures due to examples in this repo being outdated (old `@context`)**

- Add [nipoppy/pipeline-catalog](https://github.com/nipoppy/pipeline-catalog) as submodule
- Create new `bagel derivatives` command
  - Add basic input checks of processing status TSV: no missing subject IDs, no new (unique) subjects, only recognized pipeline names + vers
  - Add command logic to extract completed pipelines per subject-session based on step statuses
  - Aggregate valid pipeline info for missing sessions for a subject under a new custom `ses-nb01` ImagingSession
- Move I/O utils shared between commands into a new module `file_utils.py` to avoid circular imports
- Add new `derivative_utils.py` to hold helper funcs for `bagel derivatives`
- Add new example processing status TSVs (7 total)
- Switch to re-generating the `@context` in each command (previously, we were stripping it from an input JSONLD and re-adding it at the end. This led to sometimes ending up with an incomplete `@context` without all the terms used/added to the dataset by the command that was just run)

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


bagel/utility.py
surchs, line 1: every other file in this category is called `utils`. Want to call it `utils.py` as well?

bagel/utility.py
```diff
@@ -1,49 +1,78 @@
...
+) -> list:
+    """Check a list of subject IDs and return any not found in the provided phenotypic subject list."""
+    return list(set(subjects).difference(pheno_subjects))
+
+
+def extract_and_validate_jsonld_dataset(
+    jsonld: dict, file_path: Path
+) -> models.Dataset:
+    """
+    Strip the context from a user-provided JSONLD and validate the remaining contents
```
surchs, line 48: why do you need a path and a jsonld dictionary input here? Why can't this function just call the `load_json` function directly with the Path? it seems the only use of `jsonld` is to be fed to this function here.

bagel/tests/test_utility.py
```diff
@@ -8,15 +7,21 @@
...
+import bagel.file_utils as futil
 import bagel.pheno_utils as putil
-from bagel import mappings
-from bagel.utility import load_json
+from bagel import mappings, models
+from bagel.utility import (
+    extract_subs_from_jsonld_dataset,
+    generate_context,
+    get_subs_missing_from_pheno_data,
+)
```
surchs, line 19: very minor but why do we namespace import all of the other util modules, but expand the `bagel.utility` imports?

bagel/tests/test_utility.py
```diff
@@ -352,25 +357,29 @@ def test_invalid_age_heuristic():
     assert "unrecognized age transformation: nb:birthyear" in str(e.value)
 
 
+# TODO: See if we can remove this test: it's a little hard to maintain and
+# essentially replicates the logic of the function being tested
+# Instead, see test_all_used_namespaces_have_urls and test_used_namespaces_in_context
 @pytest.mark.parametrize(
     "model, attributes",
```
surchs, line 364: Good point! Could you make an issue for that?

bagel/tests/test_utility.py
```diff
@@ -379,6 +388,23 @@ def test_invalid_age_heuristic():
...
+        ),
+        (
+            "Subject",
+            [
+                "hasLabel",
+                "hasSession",
+                "schemaKey",
+            ],
+        ),
         (
```
surchs, line 408: Did we miss this before?

bagel/tests/test_utility.py
```diff
@@ -414,34 +440,39 @@ def test_get_bids_subjects_simple(bids_path, bids_dir):
...
+    # We sort the list for comparison since the order of the missing subjects is not guaranteed
+    # due to using set operations
+    assert (
+        sorted(
+            get_subs_missing_from_pheno_data(
+                pheno_subjects=pheno_list, subjects=bids_list
+            )
         )
+        == missing_subs
+    )
```
surchs, line 476: Let's stick with the classic test structure of: setup, act, assert. Better for readability and for making changes later.

bagel/tests/test_utility.py
```diff
@@ -543,7 +574,216 @@ def test_failed_json_reading_raises_informative_error(
...
+
+def test_extract_subs_from_jsonld_dataset(
+    test_data_upload_path, load_test_json
+):
+    """Test that subjects are correctly extracted from a JSONLD dataset."""
+    dataset = load_test_json(
+        test_data_upload_path / "example_synthetic.jsonld"
+    )
+    dataset.pop("@context")
+    subjects = extract_subs_from_jsonld_dataset(
```
surchs, line 591: Why not just make a models.Dataset instance here with some known subjects and then feed that in? 
My concern is that we're now testing two functions (however simple), and you assert over "is it a subject instance"
instead of "do we have the right subject labels" which is arguable more interesting (but obscured by loading the external file).

bagel/tests/test_utility.py
```diff
@@ -543,7 +574,216 @@ def test_failed_json_reading_raises_informative_error(
...
+    ],
+)
+def test_unrecognized_pipeline_names_raise_error(pipelines, unrecog_pipelines):
+    """Test that pipeline names not found in the pipeline catalog raise an informative error."""
+    with pytest.raises(LookupError) as e:
+        dutil.check_pipelines_are_recognized(pipelines)
+
+    assert all(
+        substr in str(e.value)
+        for substr in ["unrecognized pipelines"] + unrecog_pipelines
```
surchs, line 635: wow, nifty

bagel/tests/test_utility.py
```diff
@@ -543,7 +574,216 @@ def test_failed_json_reading_raises_informative_error(
...
+                        },
+                        "schemaKey": "Acquisition",
+                    },
+                ],
+                "schemaKey": "ImagingSession",
+            },
+        ],
+        "schemaKey": "Subject",
+    }
+    example_subject = models.Subject(**example_subject_jsonld)
```
surchs, line 706: I like this a lot better than loading external files - even if it's verbose (or because it is). Very nice to read and understand.

bagel/tests/test_utility.py
```diff
@@ -543,7 +574,216 @@ def test_failed_json_reading_raises_informative_error(
...
+        )
+
+    assert all(
+        substr in str(e.value)
+        for substr in ["unrecognized fmriprep versions"] + unrecog_versions
+    )
+
+
+def test_get_subject_imaging_sessions():
+    """Test that get_subject_imaging_sessions() accurately finds imaging sessions for a given subject."""
```
surchs, line 662:
```suggestion
    """Test that get_subject_imaging_sessions() correctly matches existing imaging sessions for a given subject."""
```

bagel/tests/test_utility.py
```diff
@@ -543,7 +574,216 @@ def test_failed_json_reading_raises_informative_error(
...
+            },
+        ],
+        "schemaKey": "Subject",
+    }
+    example_subject = models.Subject(**example_subject_jsonld)
+
+    assert list(
+        dutil.get_subject_imaging_sessions(example_subject).keys()
+    ) == ["ses-im01"]
+
```
surchs, line 711: Same as before, let's separate the "act" and the "assert" step

bagel/tests/test_utility.py
```diff
@@ -543,7 +574,216 @@ def test_failed_json_reading_raises_informative_error(
...
+            "session_id",
+            "bids_session",
+            "pipeline_name",
+            "pipeline_version",
+            "pipeline_step",
+            "status",
+        ],
+        data=sub_ses_data,
+    )
+    completed_pipelines = dutil.create_completed_pipelines(example_ses_proc_df)
```
surchs, line 764: nice

bagel/tests/test_utility.py
```diff
@@ -543,7 +574,216 @@ def test_failed_json_reading_raises_informative_error(
...
+    based on completion status of all pipeline steps.
+    """
+    sub_ses_data = [
+        [
+            "01",
+            "sub-01",
+            "01",
+            "ses-01",
+            "fmriprep",
+            "20.2.7",
```
surchs, line 725: I don't get this? Why does your fmriprep have a different version across all steps? 
Are you testing that only the version tag on the last step is read out? If so, you should definitely make that explicit. 
Otherwise I would make all fmriprep versions of the same subject/session have the same version across steps. Because that's what I would expect

bagel/tests/test_utility.py
```diff
@@ -543,7 +574,216 @@ def test_failed_json_reading_raises_informative_error(
...
+            "01",
+            "ses-01",
+            "fmriprep",
+            "23.1.3",
+            "default",
+            "SUCCESS",
+        ],
+    ]
+    example_ses_proc_df = pd.DataFrame.from_records(
+        # TODO: Don't hardcode col names?
```
surchs, line 751: That's fine, and makes the table easier to understand. In fact, you could instead setup the table as a dictionary and make it even easier to read

bagel/tests/test_utility.py
```diff
@@ -543,7 +574,216 @@ def test_failed_json_reading_raises_informative_error(
...
+        data=sub_ses_data,
+    )
+    completed_pipelines = dutil.create_completed_pipelines(example_ses_proc_df)
+
+    assert len(completed_pipelines) == 1
+    assert (
+        completed_pipelines[0].hasPipelineName.identifier
+        == f"{mappings.NP.pf}:fmriprep"
+    )
+    assert completed_pipelines[0].hasPipelineVersion == "23.1.3"
```
surchs, line 771: as mentioned above, assertion is a little hard to understand here since there are different versions in your table. Once that's cleared up, the assertion might need a little comment to explain what's tested here

bagel/tests/test_cli_derivatives.py
```diff
@@ -0,0 +1,207 @@
...
+    runner,
+    test_data,
+    test_data_upload_path,
+    default_derivatives_output_path,
+    load_test_json,
+):
+    """
+    Test that when a valid processing status TSV and bagel pheno- and bagel bids-created JSONLD are supplied as inputs,
+    the bagel derivatives command successfully creates an output JSONLD where processing info for existing vs. new
+    sessions is handled correctly.
```
surchs, line 70:
```suggestion
    the bagel derivatives command successfully creates an output JSONLD where pipeline completion info
    is correctly added to existing or newly created imaging sessions.
```

bagel/tests/test_cli_derivatives.py
```diff
@@ -0,0 +1,207 @@
...
+            ],
+            catch_exceptions=False,
+        )
+
+    for substring in expected_message:
+        assert substring in str(e.value)
+
+    assert (
+        not default_derivatives_output_path.exists()
+    ), "The JSONLD output was still created."
```
surchs, line 154:
```suggestion
    ), "The JSONLD output was created despite inputs being invalid."
```

bagel/cli.py
```diff
@@ -335,3 +346,163 @@ def bids(
...
+    BIDS metadata (from the bagel bids command).
+    NOTE: Must be run AFTER the pheno command.
+
+    This command will create a valid, subject-level instance of the Neurobagel
+    graph data model for the combined metadata in the .jsonld format.
+    You can upload this .jsonld file to the Neurobagel graph.
+    """
+    futil.check_overwrite(output, overwrite)
+
+    space = 51
```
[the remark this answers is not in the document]
alyssadai: yeah, exactly: `space` is used to left-align the file paths in the following `print` statement - purely for aesthetic purposes. It determines the number of characters that should be taken up by everything before the path, with spaces to fill in characters as needed.

It achieves something that looks like this:
```
Running initial checks of inputs...
   Existing subject graph data to augment (.jsonld):  /some/path/to/jsonld
   Processing status file (.tsv):                     /some/path/to/tsv
```

The way I get `space` for each CLI command is pretty manual: I take the char length of the longer file description (in this case, `Existing subject graph data to augment (.jsonld):` + 3 leading spaces) and tack on two more characters for after the `:` = 51.

I can add a comment explaining this for now, and we can create a little util for formatting these print statements in the future if you think that's worthwhile!

bagel/cli.py
```diff
@@ -254,25 +264,25 @@ def bids(
...
 
-    pheno_subject_dict = {
-        pheno_subject.hasLabel: pheno_subject
-        for pheno_subject in getattr(pheno_dataset, "hasSamples")
-    }
+    pheno_subject_dict = extract_subs_from_jsonld_dataset(pheno_dataset)
 
     # TODO: Revert to using Layout.get_subjects() to get BIDS subjects once pybids performance is improved
-    butil.check_unique_bids_subjects(
+    unique_bids_subjects = get_subs_missing_from_pheno_data(
```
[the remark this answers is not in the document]
alyssadai: I can address this in https://github.com/neurobagel/bagel-cli/issues/346 since that will include other functional changes to the bids command.

bagel/derivatives_utils.py
```diff
@@ -0,0 +1,89 @@
...
+        ]
+    ):
+        # Check that all pipeline steps have succeeded
+        if (
+            session_pipe_df[PROC_STATUS_COLS["status"]].str.lower()
+            == "success"
+        ).all():
+            completed_pipeline = models.CompletedPipeline(
+                hasPipelineName=models.Pipeline(
+                    identifier=mappings.get_pipeline_uris()[pipeline]
```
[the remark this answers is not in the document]
alyssadai: Ok. Could you remind me again why we don't want to run + store the output of `get_pipeline_uris()` and `get_pipeline_versions()` in mappings.py in variables like 
```
KNOWN_PIPELINES = get_pipeline_uris()
KNOWN_VERSIONS = get_pipeline_uris()
```
and then just import these variables into other modules?
Since these funcs just rely on the submodule, the catalog files should essentially always be available when a user runs the CLI, no?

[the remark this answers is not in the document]
alyssadai: Also, in this specific function, it's only used once, so I'm not sure it's worth storing it in a separate local variable.

bagel/tests/test_utility.py
alyssadai: yes

bagel/utility.py
alyssadai: will address in https://github.com/neurobagel/bagel-cli/issues/350 !

bagel/cli.py
```diff
@@ -329,9 +339,169 @@ def bids(
...
+                )
+                jsonld_subject.hasSession.append(new_img_session)
+
+    context = generate_context()
+    merged_dataset = {**context, **jsonld_dataset.dict(exclude_none=True)}
+
+    with open(output, "w") as f:
+        f.write(json.dumps(merged_dataset, indent=2))
+
+    print(f"Saved output to:  {output}")
```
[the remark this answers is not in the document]
alyssadai: will address in https://github.com/neurobagel/bagel-cli/issues/350

bagel/cli.py
```diff
@@ -335,3 +346,163 @@ def bids(
...
+        status_df, [PROC_STATUS_COLS["participant"]]
+    ):
+        raise LookupError(
+            f"Your processing status file contains missing values in the column '{PROC_STATUS_COLS['participant']}'. "
+            "Please ensure that every row has a non-empty participant id. "
+            f"We found missing values in the following rows (first row is zero): {row_indices}."
+        )
+
+    pipelines = status_df[PROC_STATUS_COLS["pipeline_name"]].unique()
+    dutil.check_pipelines_are_recognized(pipelines)
```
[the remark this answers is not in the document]
alyssadai: I think I prefer the current approach because the logic is more explicit, the individual function has fewer things to worry about + a simpler input, and thus is also easier to test. At the moment, we need to extract the unique pipeline names anyways to check that their corresponding versions in the TSV are also recognized, via `check_pipeline_versions_are_recognized()`. We could combine all these steps into a single validation function, but again, I think this might sacrifice test maintainability/readability as well as modularity.

If you have a strong preference for refactoring out this validation entirely, maybe we can open a separate issue to discuss it?

bagel/utility.py
alyssadai: good call!

bagel/tests/test_utility.py
alyssadai: will address as part of https://github.com/neurobagel/bagel-cli/issues/350!

bagel/cli.py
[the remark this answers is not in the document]
surchs: Thanks for the explainer! No strong feelings on the comment, maybe "51 seems to work well for usual path lengths to look nice" helps the next reader. But very optional

bagel/cli.py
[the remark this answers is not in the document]
surchs: agree that this would be in a separate issue - and no strong opinion on whether that's needed. Just to be clear: I'd keep the `check_pipelines_are_recognized` around. I'd just have a new `return_recognized_pipelines` function that we feed the whole `status_df` to and that internally handles looking up the right column name, getting the unique values from it, running them into check_pipelines to weed out the evil doers / or just error out, and then returns the rest. 

So essentially the same syntax you have here, same modularity, just a bit less stuff happening directly in the CLI. Should be pretty testable too

bagel/derivatives_utils.py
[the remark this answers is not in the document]
surchs: > Ok. Could you remind me again why we don't want to run + store the output of `get_pipeline_uris()` and `get_pipeline_versions()` in mappings.py in variables like
> 
> ```
> KNOWN_PIPELINES = get_pipeline_uris()
> KNOWN_VERSIONS = get_pipeline_uris()
> ```
> 
> and then just import these variables into other modules?

I guess that could be justified, yes. Generally you want to avoid side-effects when importing a module so it can be used flexibly in contexts that may e.g. not require the config. So it's a good pattern to avoid top-level code execution in modules except for constants and such. But that's more true for packages. CLI is an app, and the config is needed for the app to run. Essentially what you're saying is: if you didn't setup the pipeline-catalog submodule and you run some part of the code/app (e.g. during testing) that imports this module for whatever reason, we're OK with that crashing. And that's probably OK.

bagel/tests/test_utility.py
alyssadai: Hmm, no real reason I think, maybe just as another way to differentiate command-specific vs. common utilities when looking at the code. There tends to be fewer shared utilities, so it wasn't much additional effort to import them. That said, it probably doesn't hurt to make this consistent with the other modules. Could address in #350

bagel/cli.py
```diff
@@ -329,9 +339,169 @@ def bids(
...
+    )
+    if len(unique_derivatives_subs) > 0:
+        raise LookupError(
+            "The specified processing status file contains subject IDs not found in "
+            "the provided json-ld file:\n"
+            f"{unique_derivatives_subs}\n"
+            "Subject IDs are case sensitive. "
+            "Please check that the processing status file corresponds to the dataset in the provided .jsonld."
+        )
+
```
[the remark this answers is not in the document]
alyssadai: The main reason for this refactoring was because the error messages are command-specific. That said, I can adopt the approach we used to refactor `load_tabular()` and try to create a common error message template, and just replace file titles in the string via params. It gives us a bit less flexibility but should work for now !

bagel/tests/test_utility.py
alyssadai: >  Why does your fmriprep have a different version across all steps?

It doesn't, this example illustrates a subject-session that has run two different pipelines:
- fmriprep-20.2.7, with 2 steps (step1 - SUCCESS, step2 - FAIL)
- fmriprep-23.1.3, with 1 step (default - SUCCESS)

The goal of the test is to assert that only fmriprep-23.1.3 is interpreted by the CLI as a "completed pipeline", because **all** steps of this pipeline known to be run for the subject-session (just 1 step in this case) have a "SUCCESS" status. fmriprep-20.2.7 doesn't meet this condition because step2 failed, so it should be in the output of `create_completed_pipelines()`. I can update the docstring to be clearer.

bagel/tests/test_utility.py
alyssadai: see reply to your comment above!

bagel/tests/test_utility.py
surchs: Aha aha, I see it now. Very cool!

surchs (approved):
Thanks very much @alyssadai, looks great.

:cook:

alyssadai:
Keeping this PR open for now to merge branch for https://github.com/neurobagel/bagel-cli/issues/346 back into this one

Issue #180, opened 2023-07-21
Migrate to Pydantic v2


alyssadai:
The release of Pydantic V2 introduced several breaking changes in the code due to name changes of multiple methods, removed or deprecated types/methods/objects, and broader changes to the logic of fields, validators, + more.

The docs for V2 have _not quite_ caught up to the changes yet, but there is a comprehensive migration guide 🎉 
- https://docs.pydantic.dev/dev-v2/migration/ and pretty detailed `PydanticUserErrors`
- (new!) https://github.com/pydantic/bump-pydantic

Related are:
#179

TODO: 
- [x] investigate needed refactor for moving to pydantic 2
- [x] refactor to enable pydantic 2 and migrate

alyssadai:
Any thoughts @surchs?

I've started making some of the necessary changes (based on currently failing tests) in a branch and think the migration shouldn't require any major rewrites on our end, but some of the updates related to deprecated methods would probably benefit from a closer look as they involve a bit of extra code to preserve existing behavior.

surchs:
Bumping this up since my ill advised late night commit (https://github.com/neurobagel/bagel-cli/commit/1d4cb5217813cedfbe6d78fcb2821fac72059089) to main has made this an actual blocker.

alyssadai:
Will pin pydantic version for now. Taking this issue off of backlog in favour of #184.

PR #356, opened 2024-10-04
[ENH] Update `bagel bids` to handle JSONLD inputs incl. derivative metadata

### LINKED ISSUE #346: update `bagel bids` subcommand to play nice with new derivatives

alyssadai:
- Closes #346 

- Both `bagel bids` and `bagel derivatives` can now handle input JSONLDs which have metadata from the other command, and add data to existing imaging sessions
  - This includes existing Neurobagel-created sessions, when sessions are missing in the dataset
- `get_imaging_session_instances()` turned into a shared util

**NOTE: This PR will not have a PR label b/c it is being merged into another branch instead of `main`**

## Checklist

- [ ] PR has a label for the release changelog or `skip-release` (to be applied by maintainers only)
- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/bagel/cli.py
+++ b/bagel/cli.py
@@ -285,7 +285,9 @@ def bids(
     print("Merging BIDS metadata with existing subject annotations...\n")
     for bids_sub_id in layout.get_subjects():
         existing_subject = existing_subs_dict.get(f"sub-{bids_sub_id}")
-        session_list = []
+        existing_sessions_dict = dutil.get_imaging_session_instances(
+            existing_subject
+        )
 
         bids_sessions = layout.get_sessions(subject=bids_sub_id)
         if not bids_sessions:
@@ -302,7 +304,6 @@ def bids(
                 session=session,
             )
 
-            # If subject's session has no image files, a Session object is not added
             if not image_list:
                 continue
 
@@ -311,7 +312,10 @@ def bids(
             # so the API can still find the session-level information.
             # This should be revisited in the future as for these cases the resulting dataset object is not
             # an exact representation of what's on disk.
-            session_label = CUSTOM_SESSION_ID if session is None else session
+            # Here, we also need to add back "ses" prefix because pybids stripped it
+            session_label = "ses-" + (
+                CUSTOM_SESSION_ID if session is None else session
+            )
             session_path = butil.get_session_path(
                 layout=layout,
                 bids_dir=bids_dir,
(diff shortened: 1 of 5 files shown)


bagel/cli.py
```diff
@@ -305,36 +305,43 @@ def bids(
...
                     hasAcquisition=image_list,
                 )
-            )
-
-        pheno_subject.hasSession += session_list
+                existing_subject.hasSession.append(new_imaging_session)
 
     context = generate_context()
-    merged_dataset = {**context, **pheno_dataset.dict(exclude_none=True)}
+    merged_dataset = {**context, **jsonld_dataset.dict(exclude_none=True)}
```
surchs, line 344: My IDE says about this line:

```
The method "dict" in class "BaseModel" is deprecated
  The `dict` method is deprecated; use `model_dump` instead.Pylance
```
Mind taking a look?

bagel/cli.py
```diff
@@ -440,33 +446,28 @@ def derivatives(
...
-        # Get existing imaging sessions for the subject
-        # Note: This dictionary can be empty if only bagel pheno was run
-        existing_sessions_dict = dutil.get_imaging_session_instances(
+        # Note: Dictionary of existing imaging sessions can be empty if only bagel pheno was run
+        existing_sessions_dict = get_imaging_session_instances(
             existing_subject
         )
 
-        for session_name, sub_ses_proc_df in sub_proc_df.groupby(
+        for session, sub_ses_proc_df in sub_proc_df.groupby(
```
surchs, line 454: why did you revert `session_name` -> `session`? When I see `session` I expect `isinstance(Session)`, not `session="ses-somename"`.

bagel/cli.py
```diff
@@ -440,33 +446,28 @@ def derivatives(
...
             completed_pipelines = dutil.create_completed_pipelines(
                 sub_ses_proc_df
             )
 
             if not completed_pipelines:
                 continue
-            if session_name in existing_sessions_dict:
-                existing_img_session = existing_sessions_dict.get(session_name)
+
+            session_label = CUSTOM_SESSION_ID if session == "" else session
```
surchs, line 464: Ah, I see. I feel like this is more clearly done by reassigning `session_name` or `session_label`, rather than calling it `session` first.

surchs (approved):
Thanks @alyssadai! I left some comments and suggestions, please take a look and then good to go :cook: 

One sidenote: I think it makes sense to organize some utility functions into separate modules. But there are quite a few now, and I personally don't always know which module to go looking for a specific function. One idea could be to export methods from across the utility modules from a utils package, i.e. in the `__init__.py` file. That way at least I could import them all as `from utils import get_Something, getAnotherthing`. Another idea could be to generally reduce the number of sub-modules for utility methods to something more manageable.

bagel/cli.py
alyssadai: `model_dump` is only available with Pydantic v2! Would need to be addressed in https://github.com/neurobagel/bagel-cli/issues/180

PR #360, opened 2024-10-11
[INF] Add script to generate `nb_vocab.ttl`

### LINKED ISSUE #359: Create script to automatically generate `nb_vocab.ttl` based on `models.py`
We should have an easy way to update the `nb` vocabulary file whenver the Neurobagel graph data model is updated.

alyssadai:
- Closes #359 
- See also https://github.com/neurobagel/recipes/pull/89 for updated `nb_vocab.ttl`

- Python script added to generate nb_vocab.ttl (should end up in https://github.com/neurobagel/recipes/tree/main/vocab)
- Added README instructions for when to regenerate this file

**For reviewer:**
Any thoughts on how we could make this process more automatic? At the moment, after updating `nb_vocab.ttl`, we would have to copy it over into `recipes` manually since that's where it's used. One option could be a GH workflow to regenerate this file and put it in the right place, with maybe a `workflow_dispatch` trigger only (?) - since not every bagel-cli PR will be a data model change.

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


generate_nb_vocab_file.py
```diff
@@ -0,0 +1,54 @@
...
+        name: obj
+        for name, obj in inspect.getmembers(models)
+        if inspect.isclass(obj)
+        and issubclass(obj, pydantic.BaseModel)
+        and name not in EXCLUDE_CLASSES
+    }
+    nb_vocab = ""
+
+    for count, (name, obj) in enumerate(sorted(all_classes.items())):
+        class_entry = f"{mappings.NB.pf}:{name}\n" "    a rdfs:Class"
```
surchs, line 29:
```suggestion
        class_entry = f"{mappings.NB.pf}:{name}\n    a rdfs:Class"
```
Was the string end here intentional? Why is the `a rdfs:Class` on a newline?

generate_nb_vocab_file.py
```diff
@@ -0,0 +1,54 @@
...
+        for other_name, other_obj in all_classes.items():
+            if name != other_name and issubclass(obj, other_obj):
+                class_entry += (
+                    f";\n    rdfs:subClassOf {mappings.NB.pf}:{other_name}."
+                )
+                is_subclass = True
+                break
+
+        if not is_subclass:
+            class_entry += "."
```
surchs, line 41: This should be easier I think. Could you loop [over `obj.__bases__` ](https://docs.python.org/3/reference/datamodel.html#type.__bases__)to see if there are any that are not a BaseModel? I guess that would give you the parents, not the children. But maybe you could pre-sort the models who inherit from classes other than BaseModel and then deal with those models separately.

surchs: oh wait, I think finding the parents of `obj` *would* be enough here, no?

generate_nb_vocab_file.py
```diff
@@ -0,0 +1,54 @@
...
+
+        nb_vocab += class_entry + (
+            "\n\n" if count != (len(all_classes) - 1) else ""
```
surchs, line 44:
```suggestion
            "\n\n" if count < (len(all_classes) - 1) else ""
```
Another way of looking at this would be to store `class_entry` in an iterable and then doing a `"\n\n".join(myIter)` after the loop is done.

surchs (approved):
Thanks @alyssadai, I think this would be a good script to have.

While reviewing the code I was thinking: between the script you're adding here and the SPARQL template we have in the API repos, we're maintaining a good bit of code that mainly sticks f-strings together and checks a couple of if-statements. 

It might be time to look at some of the well established templating languages out there that are designed for such complex cases, e.g. Jinia2: https://github.com/pallets/jinja/

:cook:

Issue #350, opened 2024-09-28
Refactor utility modules and test_utility.py


alyssadai:
We have a growing number of *_utils.py modules for organizing utility functions shared across or specific to CLI commands. Currently, all these utility modules live in the same top-level directory as the main app (cli.py).

Meanwhile, we have a single test suite for all util functions, which is getting pretty big.

To better organize these util functions and their tests, we should:

- [ ] create a subdirectory called `utility` or `utilities` to store all *_utils.py files
- [ ] split `test_utility.py` into util-specific test modules

alyssadai:
@surchs, does the issue description make sense to you?

surchs:
Looks good @alyssadai. So the goal would be to have a single module `util` to import from, so that any util function could be imported like so `from util import my_util_fun`, yes?

alyssadai:
> Looks good @alyssadai. So the goal would be to have a single module util to import from, so that any util function could be imported like so from util import my_util_fun, yes?

Actually, I would prefer to still import each set of utilities separately, for better organization and clarity on where functions are defined.

At the moment, utility functions live in different `.py` files based on what CLI command they are used in/relevant for: `pheno_utils.py` (for functions used by `bagel pheno`), `derivatives_utils.py`, etc. And then there are two sets of utilities shared across commands, `utility.py` (to be renamed `model_utils.py` for clarity) and `file_utils.py`.

I would have a directory structure like:
```bash
bagel/
   utils/
     __init__.py
     pheno_utils.py
     bids_utils.py
     ...
   ...
```

and then import them like:
```bash
from utils import pheno_utils as putils
```

since usually all the utilities used by a CLI command are used together. Importing utility functions one at a time would be unnecessarily verbose, I think, and make it harder to see from a given part of code where the function is defined.

PR #373, opened 2024-11-04
[FIX] Make `pipeline-catalog` submodule available to built Docker image

### LINKED ISSUE #372: [BUG] Unrecognized pipelines when adding imaging derivatives

### Expected Behavior

`fmriprep`, `freesurfer`, and `mriqc` are recognized as valid pipelines since they are in the [Nipoppy pipeline catalog](https://github.com/nipoppy/pipeline-catalog/tree/main/processing)

### Current Behavior

Getting an error. Seems like `KNOWN_PIPELINE_URIS` is empty -- perhaps the container doesn't include the pipeline catalog?

### Error message

```
│ /usr/local/lib/python3.10/site-packages/bagel/derivatives_utils.py:26 in                         │
│ check_pipelines_are_recognized                                                                   │
│                                                                                                  │
│   23 │   │   set(pipelines).difference(mappings.KNOWN_PIPELINE_URIS)                             │
│   24 │   )                                                                                       │
│   25 │   if len(unrecognized_pipelines) > 0:                                                     │
│ ❱ 26 │   │   raise LookupError(                                                                  │
│   27 │   │   │   f"The processing status file contains unrecognized pipelines in the column '    │
│   28 │   │   │   f"{unrecognized_pipelines}. "                                                   │
│   29 │   │   │   f"Allowed pipeline names are the following pipelines supported natively in N    │
│                                                                                                  │
│ ╭───────────────────────────────────── locals ──────────────────────────────────────╮            │
│ │              pipelines = array(['fmriprep', 'freesurfer', 'mriqc'], dtype=object) │            │
│ │ unrecognized_pipelines = ['fmriprep', 'freesurfer', 'mriqc']                      │            │
│ ╰───────────────────────────────────────────────────────────────────────────────────╯            │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
LookupError: The processing status file contains unrecognized pipelines in the column 'pipeline_name': ['fmriprep', 'freesurfer', 
'mriqc']. Allowed pipeline names are the following pipelines supported natively in Nipoppy (https://github.com/nipoppy/pipeline-catalog): 
{}
```

### Environment

`bagelcli v0.3.1`

### How to reproduce

Use PPMI derivatives files I shared separately

### Anything else?

_No response_

alyssadai:
- Fixes #372 

Changes proposed in this PR:

- Fixes our GH actions workflows for building + pushing images to DockerHub to check out submodules

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


.github/workflows/build_docker_nightly.yml
```diff
@@ -12,6 +12,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          submodules: recursive
```
surchs, line 16: This will pull all submodules, i.e. neurobagel-examples, bids-examples, and pipelines. Technically we only need the pipelines. It's only ~70MB (mainly for bids-examples), but that's ~ 1/3 of the whole image. Our other images are all a little large - the CLI is small by comparison. All that is to say: maybe it's worth to update the Dockerfile to not copy all the submodules in.

surchs (approved):
looks mostly good, just a :cherries: note on inflated image size 

:cook:

alyssadai:
Hey Seb, re-requesting your review here for an extra pair of eyes on the Dockerfile changes.

Agree with your suggestion to only copy in required files. As part of maybe https://github.com/neurobagel/bagel-cli/issues/350, I think we can also move the test suite out of the `bagel` directory for the same reason, since we don't need to copy the tests into the image (and it's best practice to keep them separate from the main program code, like how we're doing in the API repos).

Dockerfile
```diff
@@ -2,7 +2,10 @@
 FROM python:3.10-slim-buster
 
 WORKDIR /app
-COPY . /app/src
+COPY requirements.txt pyproject.toml setup.cfg setup.py /app/src/
```
surchs, line 5: This works but I think it'd be cleaner and more expensive to use the docker ignore file to exclude what we don't want, rather than making that implicit by adding everything we do want

Issue #72, opened 2023-02-16
Use informative names for examples used in tests


alyssadai:
- [ ] consider re-using a few informatively named example .json + .tsv's
- [ ] use more "human understandable" syntax for referring to subjects and session in example data (see https://github.com/neurobagel/bagelbids/pull/62#discussion_r1106234153)

PR #379, opened 2024-11-06
[FIX] Package `pipeline-catalog` as data files in CLI installation

### LINKED ISSUE #375: [BUG] Pipeline catalog is now inside the container but still not found by the CLI

### Expected Behavior

_No response_

### Current Behavior

- Inside the container, the pipeline catalog is in `/app/src/pipeline-catalog`, thanks to #373
- This does not match the value in `bagel.mappings.PROCESSING_PIPELINE_PATH`, which points to `/usr/local/lib/python3.10/site-packages/pipeline-catalog/processing`
- This is because `PROCESSING_PIPELINE_PATH` is built from the file at `/usr/local/lib/python3.10/site-packages/bagel/mappings.py`, which is part of the **installed** Python package, not the source code
- So I think this is more of a packaging problem, not a Docker-building problem. A solution would be to move the `pipeline-catalog` submodule inside the `bagel` directory. Or there could be a script that generates a static JSON file from the pipeline catalog and stores it inside the `bagel` directory, then at runtime that JSON file could be loaded into `KNOWN_PIPELINE_URIS` directly. A third possible solution would be to configure `pyproject.toml` to somehow include the pipeline catalog, but I'm not sure how to do that and also that might not guarantee that `PROCESSING_PIPELINE_PATH` will still point to the correct directory.

### Error message

_No response_

### Environment

`bagelcli v0.3.2`

### How to reproduce

_No response_

### Anything else?

_No response_

### LINKED ISSUE #376: CI should not install in dev mode

### Expected Behavior

My CI tests run using the same environment I would expect to have when using the tools as a user. 

### Current Behavior

The CI runs our tests after install the tool in [dev-mode](https://pip.pypa.io/en/latest/topics/local-project-installs/#editable-installs):

https://github.com/neurobagel/bagel-cli/blob/c4a4616a7578e1ca15b2c9ab4a738ceb15574b29/.github/workflows/test.yml#L26

For the docker image, we build things correctly

https://github.com/neurobagel/bagel-cli/blob/c4a4616a7578e1ca15b2c9ab4a738ceb15574b29/Dockerfile#L11

### Error message

The main issue here seems to be that relative paths (e.g. to submodules in the repo) are resolving correctly
when the tests run in dev-mode, but they no longer resolve correctly when the installed regularly.

This is currently causing all kinds of issues around the newly introduced submodule dependencies. but the 
underlying problem is that we test in a different install mode than we run our tools

### Environment

- OS:
- Python/Node version:

### How to reproduce

_No response_

### Anything else?

Related to #375

alyssadai:
- Fixes #375
- Fixes #376 

- `pipeline-catalog` submodule moved into `bagel` module so that it ends up inside the package directory after installation
  - `pipeline-catalog` wouldn't be automatically included in the installation since it's not a module (no `__init__.py`), so we additionally must update the package config file to include this subdirectory as additional package data
- for separation of concerns, test suite moved out of `bagel` module, and test data submodules (`neurobagel_examples`, `bids-examples`) moved under `tests/`
- for CI tests, CLI is no longer installed in editable mode to more closely replicate production env

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


.gitmodules
```diff
@@ -1,9 +1,9 @@
 [submodule "bids-examples"]
-	path = bids-examples
+	path = tests/bids-examples
```
surchs, line 2: why not `tests/data`? I would say that's where this fits better, no?

tests/test_cli_bids.py
```diff
@@ -133,6 +133,7 @@ def test_imaging_sessions_have_expected_metadata(
 def test_bids_data_with_sessions_have_correct_paths(
     runner,
     test_data_upload_path,
+    bids_synthetic,
```
surchs, line 136: did we just forget to use this before?

surchs (approved):
Thanks @alyssadai for fixing this. 

I understand that my guess of dev-mode install didn't fix it, so having a test here that would have caught the original bug is tricky. So let's do this: let's merge the fix without the test, but please create an issue to find a way to run the tests on the installed package. My guess is that the CLIRunner we use to "call" the CLI as if we used it from the command line is the limitation here, but I'm sure there is a way to force it to run on the installed CLI (even something silly like making an actual system call with subprocess or something - but I'm sure something smarter must exist too).

In any case: thanks for the fix :cook:

tests/test_cli_bids.py
alyssadai: I think so!

.gitmodules
alyssadai: Fair point - I thought about this too, but I'm not sure if I 100% agree... `tests/data` includes files that we maintain purely for testing the CLI, and every file in there is used in the test suite (and should be documented in `test/data/README.md`).

In contrast, we only use a few files from `bids-examples` and `neurobagel-examples`, and since they are external repos, I think it makes sense to keep them separate from the static TSV & JSON test input files.

I think an option that might make more sense is something like:

```bash
tests/
   data/
     bids-examples/
     neurobagel_examples/
     tsv_json_datasets/  # or similar
   ...
```

with everything currently in `tests/data` moved to `tsv_json_datasets`. But I'm honestly not sure that the extra layer of abstraction is needed or that much better than what we currently have, since we already have fixtures for each of these paths anyhow. 🤷

From a visibility / submodule tracking standpoint, I'm also a little hesitant to have the submodules buried too deep in the directory tree 😅

For now, I've added this as a checkbox to consider in our other existing issue for reorganizing the actual test data: https://github.com/neurobagel/bagel-cli/issues/72.

PR #385, opened 2024-11-07
[REF] Refactor utilities

### LINKED ISSUE #164: Refactor the tests to be clearer and better differentiated between integration and unit
We now have a lot of example test files with very minimal differences between them. That is because we are running most of our tests as integration tests (e.g. run through the entire app and looking at it from the outside). That makes sense to make sure that certain input files lead to certain behaviour of the entire app. I don't think this makes as much sense for asserting that very specific error messages are raised or warnings given. For this, we could just use a unit test of the function that does the raising or warning.

In return, we might be able to condense our examples to a couple of "valid" ones with some allowed variability, a couple of ones that should warn, and a couple that should fail outright. it will also make understanding the tests easier and be clearer to reason about if and when these tests fail.

Other potential options:
- order the tests in some semantically meaningful way / split them into more test modules so we know where to look for and where to add similar tests
- set up pytest markers for e.g., integration vs. unit tests following https://docs.pytest.org/en/7.1.x/how-to/mark.html

### LINKED ISSUE #350: Refactor utility modules and test_utility.py

alyssadai:
- Closes #350
- Closes #164 

- move all utility files into a `utilities` module
- rename `utility.py` to `model_utils.py` (includes utility functions related to generating + extracting entities from graph dataset instances, etc.)
- split `test_utility.py` into:
  - `test_putil.py`
  - `test_butil.py`
  - `test_dutil.py`
  - `test_futil.py`
  - `test_mutil.py`
  - and organized them into `integration/` and `unit/` subdirs
- remove test `test_generate_context()` (it replicates logic of function being tested and is hard to maintain)
- refactor out into utility functions:
  - merging context + final dataset instance for output to JSONLD
  - saving final JSONLD
 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


bagel/cli.py
```diff
@@ -1,22 +1,16 @@
...
-    generate_context,
-    get_imaging_session_instances,
-    get_subject_instances,
-)
+
+from .utilities import bids_utils as butil
+from .utilities import derivatives_utils as dutil
+from .utilities import file_utils as futil
+from .utilities import model_utils as mutil
+from .utilities import pheno_utils as putil
```
surchs, line 12: :cherries: is that necessary? I find `file_utils` clearer to understand than `futil` etc as it also matches the module name. Might make the code a little longer though ...

surchs: alternative idea: package `utilities` -> `util` and then you can just do `import util` and refer to `util.file`, `util.bids`, `util.model`. almost as short, but still clear

bagel/utilities/model_utils.py
```diff
@@ -5,10 +5,11 @@
 import typer
 from pydantic import ValidationError
 
-import bagel.file_utils as futil
 from bagel import models
 from bagel.mappings import ALL_NAMESPACES, NB
 
+from . import file_utils as futil
```
surchs, line 11:
```suggestion
from utilities import file_utils as futil
```

tests/unit/test_butil.py
surchs, line 1: :cherries: I think test name can be little longer to be clear. i.e. `test_bids_util`. Alternatively you could have a subdiirectory `unit/utilities/test_bids.py` to make it clear like that. If it has to be short, I wouldn't save space on the `b(ids)` part, rather on the repeating `u(tils)` part to keep this clear - especially because I don't have these modules sitting right next to the tests in the source tree anymore

surchs (approved):
Thanks @alyssadai for this refactor. I think it looks very nice! 
I have some very minor clarity comments, take a look.

> remove test test_generate_context() (it replicates logic of function being tested and is hard to maintain)

agreed, but I think we already had the case that we didn't include a class in the context and that then breaks things in confusing ways. So I don't think removing the test completely is a solution either. I don't have a great idea, but let's discuss this in a follow-up issue.

:cook:

bagel/utilities/model_utils.py
alyssadai: Pytest fails with this change, seemingly because `utilities` is not recognized as a top-level module, but rather one nested inside `bagel`. The relative import seems helpful here for keeping the imports self-contained from inside the `bagel` package.

surchs: Aha right, my bad. It would have to be `bagel.utilities` of course.

alyssadai:
Hey @surchs, just to reply to your concern regarding the removed test:

> I think we already had the case that we didn't include a class in the context and that then breaks things in confusing ways. So I don't think removing the test completely is a solution either

To clarify, we added two new tests in PR #349, `test_all_used_namespaces_have_urls` and `test_used_namespaces_in_context`, that should catch these issues without requiring us to remember to update the test parameters whenever we update the data model! :) The removed test was just leftover from a `#TODO` in that PR.

PR #391, opened 2024-11-15
[ENH] Load pipeline catalog dynamically and error out when it is empty/not found

### LINKED ISSUE #374: Error out when `pipeline-catalog` not found or returns empty index of pipelines
As seen in https://github.com/neurobagel/bagel-cli/issues/372, because CLI functionality now relies on an external repo submodule (`pipeline-catalog`), we can run into scenarios where the submodule (for whatever reason) was not initialized properly, leading to an incomplete index of known pipelines and versions.

This then causes the `derivatives` command to fail because none of the pipeline names in a processing status TSV are recognized, even if they are technically valid.

To catch this issue, we should error out & exit immediately in the CLI when fetching allowed pipelines returns an empty array:

https://github.com/neurobagel/bagel-cli/blob/66111d49d468e3debe20645ac083b47e47aa153c/bagel/mappings.py#L46-L73

### LINKED ISSUE #383: `pipeline-catalog` should be fetched dynamically
We should switch to grabbing https://github.com/nipoppy/pipeline-catalog/blob/main/processing/processing.json from `pipeline-catalog` via a GET request.

The submodule should be used as a backup in case the request fails, similar to how we're using the Cognitive Atlas in the API. 

This prevents the need to re-release whenever there is an update to `pipeline-catalog`.

Related to: https://github.com/nipoppy/pipeline-catalog/issues/5

surchs:
No longer needed

- Closes #383
- Replaces #386
- Closes #374

- new getter function that uses `requests` to load the catalog
- failing that, falls back to a local copy of the catalog (that we'll need to keep up to date)

**For reviewer**: seems like mocking `requests` is a whole thing. My unit test mainly ensures that our fallback local file works if the request to the GitHub pipeline catalog fails. But we may want to be a bit smarter about testing this in the future. Maybe you have ideas.

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

## Summary by Sourcery

Enhance the CLI to dynamically load the pipeline catalog from a remote source, with a fallback to a local copy if the remote is unavailable. Update the project dependencies to include 'httpx' for HTTP requests and add tests to ensure the robustness of the catalog loading process.

Enhancements:
- Implement dynamic loading of the pipeline catalog in the CLI, allowing it to fetch the catalog from a remote source and fall back to a local copy if the remote is unavailable.

Build:
- Add 'httpx' and its dependencies to the project requirements for handling HTTP requests.

Tests:
- Add tests to ensure the pipeline catalog is loaded from a local backup if the remote source is unreachable, and raise an exception if both sources fail.
- Add tests to verify successful loading of the pipeline catalog from a remote source.


surchs:
@alyssadai this should be ready for you now

tests/unit/test_derivative_utils.py
```diff
@@ -1,27 +1,110 @@
...
+    # or switch to httpx for better testing capabilities
+    nonsense_url = "https://does.not.exist.url"
+
+    def mock_httpx_get(*args, **kwargs):
+        response = httpx.Response(status_code=400, json={"key": "value"})
+        # This slightly odd construction is necessary to create a Response object
+        # that has the correct URL already baked in (I think), because otherwise we get the following
+        # RuntimeError: Cannot call `raise_for_status` as the request instance has not been set on this response.
+        # TODO: find a better solution or understand the problem better
+        response._request = httpx.Request("GET", nonsense_url)
```
alyssadai, line 23: I think this makes sense to me: I guess `raise_for_status` is dependent on a `Request` object being available to be able to generate an exception.

I think your solution looks fine, but 🍒 I would propose not replacing `response._request` post-hoc, and instead defining it where you define the `response` variable:

```python
response = httpx.Response(
    status_code=400, json={"key": "value"}, request=httpx.Request("GET", nonsense_url)
)
```

I find this easier to understand.

tests/unit/test_derivative_utils.py
```diff
@@ -1,27 +1,110 @@
...
+        return response
+
+    monkeypatch.setattr(httpx, "get", mock_httpx_get)
+
+    with pytest.warns(UserWarning) as w:
+        result = mappings.get_pipeline_catalog(
+            get_url=nonsense_url, get_path=mappings.PROCESSING_PIPELINE_PATH
```
alyssadai, line 30: worth adding a note here that this test will fail if the submodule is not correctly initialized? since the test directly relies on the submodule

tests/unit/test_derivative_utils.py
```diff
@@ -1,27 +1,110 @@
...
+        )
+
+    assert all(isinstance(item, dict) for item in result)
+    assert "Unable to load pipeline catalog" in w[0].message.args[0]
```
alyssadai, line 34:
```suggestion
    assert "Unable to request pipeline catalog" in w[0].message.args[0]
```

tests/unit/test_derivative_utils.py
```diff
@@ -1,27 +1,110 @@
...
+
+
+def test_raises_exception_if_remote_and_local_pipeline_catalog_fails(
+    monkeypatch, tmp_path
+):
+    """
+    If I cannot get the pipeline catalog from the remote location and the local backup, I should raise an exception.
+    """
+    # or switch to httpx for better testing capabilities
+    nonsense_url = "https://does.not.exist.url"
```
alyssadai, line 44:
```suggestion
    nonsense_url = "https://does.not.exist.url"
```

tests/unit/test_derivative_utils.py
```diff
@@ -1,27 +1,110 @@
...
+
+    def mock_httpx_get(*args, **kwargs):
+        response = httpx.Response(status_code=400, json={"key": "value"})
+        # This slightly odd construction is necessary to create a Response object
+        # that has the correct URL already baked in (I think), because otherwise we get the following
+        # RuntimeError: Cannot call `raise_for_status` as the request instance has not been set on this response.
+        # TODO: find a better solution or understand the problem better
+        response._request = httpx.Request("GET", nonsense_url)
```
alyssadai, line 52: 🍒 as above, how about:

```suggestion
        response = httpx.Response(status_code=400, json={"key": "value"}, request=httpx.Request("GET", nonsense_url))
        # This slightly odd construction is necessary to create a Response object
        # that has the correct URL already baked in (I think), because otherwise we get the following
        # RuntimeError: Cannot call `raise_for_status` as the request instance has not been set on this response.
        # TODO: find a better solution or understand the problem better
```

tests/unit/test_derivative_utils.py
```diff
@@ -1,27 +1,110 @@
...
+
+def test_get_pipeline_from_remote_succeeds(monkeypatch):
+    nonsense_url = "https://made.up.url/pipeline_catalog.json"
+    mock_pipeline_catalog = [
+        {"name": "sillypipe", "versions": ["1", "2"]},
+        {"name": "funpipe", "versions": ["10", "11"]},
+    ]
+
+    def mock_httpx_get(*args, **kwargs):
+        response = httpx.Response(status_code=200, json=mock_pipeline_catalog)
```
alyssadai, line 73: same as above, I think we could use the `request=` param here instead of overriding it post-hoc.

tests/unit/test_derivative_utils.py
```diff
@@ -1,27 +1,110 @@
...
+        response._request = httpx.Request("GET", nonsense_url)
+        return response
+
+    monkeypatch.setattr(httpx, "get", mock_httpx_get)
+
+    result = mappings.get_pipeline_catalog(
+        get_url=nonsense_url, get_path=mappings.PROCESSING_PIPELINE_PATH
+    )
+
+    assert all(isinstance(item, dict) for item in result)
```
alyssadai, line 87: is this needed given that you check for equality of the `result` obj below?

alyssadai (commented):
Thanks for the PR @surchs!

The overall logic of the changes makes sense to me. I left some suggestions and a few comments for code clarity, and I think there's one potential modification to be made to ensure we don't emit an irrelevant error message by using one of the existing CLI utilities.

See what makes sense to you!

bagel/mappings.py
```diff
@@ -40,35 +44,76 @@
...
+def get_pipeline_catalog(get_url: str, get_path: Path) -> dict:
+    """
+    Load the pipeline catalog from the remote location or, if that fails,
+    from the local backup.
+    """
+    try:
+        response = httpx.get(get_url)
+        response.raise_for_status()
+        return response.json()
+    except (httpx.HTTPStatusError, httpx.ConnectError, json.JSONDecodeError):
```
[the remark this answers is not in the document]
surchs: Good call, I picked the exceptions because my tests triggered them . But I agree, let's handle the more generic types here

bagel/mappings.py
```diff
@@ -40,35 +44,76 @@
...
+        warnings.warn(
+            f"Unable to load pipeline catalog from {get_url}.\n"
```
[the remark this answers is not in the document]
surchs: Download? Agree that load isn't precise but request isn't user friendly

bagel/mappings.py
```diff
@@ -40,35 +44,76 @@
...
+            f"Will revert to loading backup from {get_path}."
+        )
+        try:
+            catalog = file_utils.load_json(get_path)
+            return catalog
+        except FileNotFoundError as e:
+            raise FileNotFoundError(
+                f"Unable to find a local pipeline-catalog backup. Have you correctly initialized the submodules? {e}"
```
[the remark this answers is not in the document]
surchs: If you don't have the sub modules your mouse likely doing a bare metal install or are a dev. I think you're wording still works, but we might even include a hint on how to initialize them

tests/unit/test_derivative_utils.py
surchs: Agree! Yours is much better

tests/unit/test_derivative_utils.py
surchs: Download?

surchs:
Thanks Alyssa, agree with all your comments. Unfortunately airport network is too bad for me to accept your changes, so if you'd just push directly to this branch and then accept for me, I'd appreciate!

bagel/mappings.py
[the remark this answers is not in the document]
alyssadai: Download sounds good!

bagel/mappings.py
```diff
@@ -40,35 +44,76 @@
...
+    """
+    in_arr = get_pipeline_catalog(
+        get_url=PROCESSING_PIPELINE_URL,
+        get_path=PROCESSING_PIPELINE_PATH,
+    )
+    # version_dict = {}
+    # uri_dict = {}
+    # for pipeline in in_arr:
+    #     version_dict[pipeline["name"]] = pipeline["versions"]
+    #     uri_dict[pipeline["name"]] = f"{NP.pf}:{pipeline['name']}"
```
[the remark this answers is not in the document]
alyssadai: agreed!

bagel/mappings.py
```diff
@@ -40,35 +44,76 @@
...
+        response = httpx.get(get_url)
+        response.raise_for_status()
+        return response.json()
+    except (httpx.HTTPStatusError, httpx.ConnectError, json.JSONDecodeError):
+        warnings.warn(
+            f"Unable to load pipeline catalog from {get_url}.\n"
+            f"Will revert to loading backup from {get_path}."
+        )
+        try:
+            catalog = file_utils.load_json(get_path)
```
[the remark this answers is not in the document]
surchs: Agree with the temp variable remove!

> Also, I'm wondering if it makes sense to use load_json() as-is here, since if the file is empty, it will throw a JSONDecodeError which will be caught and re-raised by the util but with a misleading error message:

Are you saying that if the file is empty or invalid, the `JSONDecodeError` will **not** be caught by the try-except, and will thus just raise on its own, which might be confusing? Since the path should be in the Error message, I don't think this is super confusing. It should be also pretty uncommon for that to happen to a regular user, as for the file to exist and also be empty/invalid, something must be wrong with our image builds - and even the imprecise error would alert us of that.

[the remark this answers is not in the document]
alyssadai: Hey @surchs, no -- I meant that the current `load_json()` utility func _will_ catch a `JSONDecodeError`, but the error message it currently always prints specifically refers to an invalid "data dictionary" (see code snippet above). This means the error message is inaccurate when the JSON being loaded is e.g., a pipeline catalog JSON, and might confuse users because the `bagel derivatives` command does not take a data dictionary as input.

Since I'm committing to this PR anyways, I'll resolve this here by updating the error message inside `load_json()` to be more generic !

alyssadai (approved):
🧑‍🍳

PR #429, opened 2025-02-06
[ENH] Add docker image build for arm64 architecture

### LINKED ISSUE #428: Cannot run CLI on Arm (e.g. Mac)

### Expected Behavior

As a user on an arm64 architecture, I can follow the same instructions as a user on x64 architecture when running the CLI docker images

### Current Behavior

We are only building and publishing images for x64, and then these images are not runnable on recent arm-based Macs.

### Error message

_No response_

### Environment

_No response_

### How to reproduce

This came up during a setup for a user on a recent Mac

### Anything else?

We should use docker multiplatform builds to also build and publish for arm, at least the tools we expect users to run locally on their machines: https://docs.docker.com/build/building/multi-platform/

surchs:
- Closes #428

- enable building CLI docker image for arm64 architecture

note: we confirmed that the new arm images are runnable on an arm mac by pulling, running and getting the help menu to show.

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Enable building and pushing of CLI Docker images for both arm64 and amd64 architectures in the CI workflow, and add a verification step to ensure the image runs correctly.

Build:
- Enable building CLI Docker image for arm64 architecture.

CI:
- Update CI workflow to build and push Docker images for both arm64 and amd64 platforms.
- Add a step in the CI workflow to run the CLI help command to verify the Docker image.

### DIFF
--- a/.github/workflows/build_docker_nightly.yml
+++ b/.github/workflows/build_docker_nightly.yml
@@ -33,3 +33,7 @@ jobs:
           file: ./Dockerfile
           push: true
           tags: ${{ secrets.DOCKERHUB_USERNAME }}/${{ vars.DOCKERHUB_REPO }}:nightly
+
+      - name: Run cli help
+        run: |
+          docker run --rm ${{ secrets.DOCKERHUB_USERNAME }}/${{ vars.DOCKERHUB_REPO }}:nightly --help
--- a/.github/workflows/build_docker_nightly.yml
+++ b/.github/workflows/build_docker_nightly.yml
@@ -35,6 +35,7 @@ jobs:
           tags: ${{ secrets.DOCKERHUB_USERNAME }}/${{ vars.DOCKERHUB_REPO }}:nightly
 
       # Verify that build was successful
+      # TODO: replace this with an actual test
       - name: Run cli help
         run: |
           docker run --rm ${{ secrets.DOCKERHUB_USERNAME }}/${{ vars.DOCKERHUB_REPO }}:nightly --help
--- a/.github/workflows/build_docker_nightly.yml
+++ b/.github/workflows/build_docker_nightly.yml
@@ -34,6 +34,7 @@ jobs:
           push: true
           tags: ${{ secrets.DOCKERHUB_USERNAME }}/${{ vars.DOCKERHUB_REPO }}:nightly
 
+      # Verify that build was successful
       - name: Run cli help
         run: |
           docker run --rm ${{ secrets.DOCKERHUB_USERNAME }}/${{ vars.DOCKERHUB_REPO }}:nightly --help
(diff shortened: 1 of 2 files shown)


.github/workflows/build_docker_nightly.yml
```diff
@@ -31,7 +28,12 @@ jobs:
       - name: Build and push
         uses: docker/build-push-action@v6
         with:
+          platforms: linux/arm64, linux/amd64
           context: .
           file: ./Dockerfile
           push: true
           tags: ${{ secrets.DOCKERHUB_USERNAME }}/${{ vars.DOCKERHUB_REPO }}:nightly
+
+      - name: Run cli help
```
alyssadai, line 37:
```suggestion
      # Verify that build was successful
      - name: Run cli help
```

.github/workflows/build_docker_nightly.yml
```diff
@@ -1,6 +1,3 @@
-# NOTE: A custom version of this template workflow is used in neurobagel/query-tool.
```
alyssadai, line 1: This comment is from the template version of this wf, synced from https://github.com/neurobagel/workflows/blob/main/template_workflows/build_docker_nightly.yml.

With this PR, this repo now also deviates from the synced template, meaning we probably should remove it from the sync.yml section for this file. I'd also suggest adding a comment about it in the template, similar to what we did for the query tool, so we don't forget.

Could you create an issue for this in https://github.com/neurobagel/workflows/issues?

alyssadai (approved):
Looks good @surchs! 🧑‍🍳 

One question: do we also plan on updating the [build_docker_on_release.yml](https://github.com/neurobagel/bagel-cli/blob/main/.github/workflows/build_docker_on_release.yml) wf? Or will we only have cross-platform support for `nightly`?

alyssadai:
@surchs, if we will also be updating [build_docker_on_release.yml](https://github.com/neurobagel/bagel-cli/blob/main/.github/workflows/build_docker_on_release.yml), I'd suggest removing the `release` label from this PR and releasing once the release wf is also updated.

surchs:
@alyssadai are you saying that you think the updated build_on_release workflow would not apply yet if we release of this PR? Or why do you suggest we update the release workflow?

alyssadai:
I meant, why not have this PR and the PR to update [build_docker_on_release.yml](https://github.com/neurobagel/bagel-cli/blob/main/.github/workflows/build_docker_on_release.yml) as part of the same release, since they are related! Instead of releasing off of this PR, and then releasing again off the next one (which will presumably update the other wf).

surchs:
Did you look at the diff? Because this PR now updates both workflows - after you pointed out that the release workflow was missing. I would prefer to do all arm related updates here, I don't see a benefit of doing the nightly and latest changes in separate PRs.

alyssadai:
@surchs ah sorry, I totally missed that commit! I think I got confused since this PR was still open. Anything else needed before we can merge?

PR #694, opened 2026-05-25
[ENH] Support catalog-mode data dictionaries

### LINKED ISSUE #686: Make the CLI compatible with new catalog-compatible data dictionaries
Now that we will introduce additional fields in data dictionaries in support of data catalog mode, the CLI needs to be updated to make sure that these new dictionaries are understood and can be validated.

### TODO

- [x] Add min/max age to the data dictionary model; ensure the CLI can handle older dictionaries that lack this field
- [x] Add total subjects to the dataset description model and validate that it is not empty
  - [x] Decide how to handle older dataset descriptions that are missing this field
- [x] Make sure that
  - [x] new type of dictionary validates
  - [x] old type of dictionary also validates
- [x] Update the data dictionary models in CLI
- [x] ignore these new values in the cli output / do not change the jsonld based on these values

surchs:
- Closes #686

- Update models for dictionary and dataset description
- Added tests for dictionary
- No tests for dataset description because we already allow additionalFields - implicit test added

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

## Summary by Sourcery

Extend pheno CLI input models to support richer dictionary and dataset description metadata.

New Features:
- Allow continuous dictionary entries to optionally specify a value range via a new ValueRange model.
- Support an optional participant_count field in dataset descriptions parsed by the CLI.

Enhancements:
- Include a new pheno example (example24) wiring dictionary, TSV, and dataset description files into the test data catalog.

Tests:
- Add an integration test ensuring the pheno CLI successfully parses catalog inputs with expanded dictionary and dataset description fields.

### DIFF
--- a/bagel/dataset_description_model.py
+++ b/bagel/dataset_description_model.py
@@ -111,14 +111,6 @@ class DatasetDescription(BaseModel):
         ),
         BeforeValidator(whitespace_string_to_default_none),
     ]
-    participant_count: Annotated[
-        int | None,
-        Field(
-            default=None,
-            description="The number of participants in the dataset",
-            alias="ParticipantCount",
-        ),
-    ]
 
     # NOTE: url_preserve_empty_path (>=2.12) is needed to prevent HttpUrl from auto-appending trailing slashes to URLs
     # when the path portion of the URL is empty
--- a/bagel/dataset_description_model.py
+++ b/bagel/dataset_description_model.py
@@ -111,6 +111,14 @@ class DatasetDescription(BaseModel):
         ),
         BeforeValidator(whitespace_string_to_default_none),
     ]
+    participant_count: Annotated[
+        int | None,
+        Field(
+            default=None,
+            description="The number of participants in the dataset",
+            alias="ParticipantCount",
+        ),
+    ]
 
     # NOTE: url_preserve_empty_path (>=2.12) is needed to prevent HttpUrl from auto-appending trailing slashes to URLs
     # when the path portion of the URL is empty
--- a/tests/data/README.md
+++ b/tests/data/README.md
@@ -30,7 +30,7 @@
 | iso88591 | invalid, ISO-8859-1 encoded and contains French characters in the age column | invalid, ISO-8859-1 encoded and contains French characters in the age column | fail |
 | invalid_json | - | not valid JSON, contains trailing comma after `group` key-value pair | fail |
 | 23 | Valid, contains multiple diagnosis, age, and sex columns | Valid | pass
-| 24 | Valid | Valid, contains ValueRange for age | pass
+| 24 | Valid | Valid, contains `ValueRange` for age and a dataset description containing `ParticipantCount` | pass
 
 `* this is expected to fail until we enable multiple participant_ID handling`.
 
--- a/tests/data/README.md
+++ b/tests/data/README.md
@@ -30,6 +30,7 @@
 | iso88591 | invalid, ISO-8859-1 encoded and contains French characters in the age column | invalid, ISO-8859-1 encoded and contains French characters in the age column | fail |
 | invalid_json | - | not valid JSON, contains trailing comma after `group` key-value pair | fail |
 | 23 | Valid, contains multiple diagnosis, age, and sex columns | Valid | pass
+| 24 | Valid | Valid, contains ValueRange for age | pass
 
 `* this is expected to fail until we enable multiple participant_ID handling`.
 
--- a/tests/data/example24.tsv
+++ b/tests/data/example24.tsv
@@ -1,0 +1,5 @@
+participant_id	session_id	group	sex	participant_age
+sub-01	ses-01	PAT	M	"P20Y6M"
+sub-01	ses-02	PAT	M	"P20Y8M"
+sub-02	ses-01	CTRL	F	"P25Y8M"
+sub-02	ses-02	CTRL	F	"P26Y4M"
--- a/tests/data/example24.tsv
+++ b/tests/data/example24.tsv
@@ -1,5 +1,5 @@
 participant_id	session_id	group	sex	participant_age
-sub-01	ses-01	PAT	M	"P20Y6M"
-sub-01	ses-02	PAT	M	"P20Y8M"
-sub-02	ses-01	CTRL	F	"P25Y8M"
-sub-02	ses-02	CTRL	F	"P26Y4M"
+sub-01	ses-01	PAT	M	P20Y6M
+sub-01	ses-02	PAT	M	P20Y8M
+sub-02	ses-01	CTRL	F	P25Y8M
+sub-02	ses-02	CTRL	F	P26Y4M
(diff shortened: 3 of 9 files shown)


alyssadai:
Hey @surchs, before I start reviewing this, could you take a look at the failing tests?

tests/data/example24.tsv
```diff
@@ -0,0 +1,5 @@
+participant_id	session_id	group	sex	participant_age
+sub-01	ses-01	PAT	M	"P20Y6M"
```
alyssadai: Just curious: why the quotes around the age values?

tests/data/README.md
```diff
@@ -30,6 +30,7 @@
 | iso88591 | invalid, ISO-8859-1 encoded and contains French characters in the age column | invalid, ISO-8859-1 encoded and contains French characters in the age column | fail |
 | invalid_json | - | not valid JSON, contains trailing comma after `group` key-value pair | fail |
 | 23 | Valid, contains multiple diagnosis, age, and sex columns | Valid | pass
+| 24 | Valid | Valid, contains ValueRange for age | pass
```
alyssadai:
```suggestion
| 24 | Valid | Valid, contains `ValueRange` for age and a dataset description containing `ParticipantCount` | pass
```

bagel/dataset_description_model.py
```diff
@@ -111,6 +111,14 @@ class DatasetDescription(BaseModel):
...
         BeforeValidator(whitespace_string_to_default_none),
     ]
+    participant_count: Annotated[
+        int | None,
+        Field(
+            default=None,
+            description="The number of participants in the dataset",
+            alias="ParticipantCount",
+        ),
+    ]
```
alyssadai: After reviewing the changes necessary to make the existing tests pass, I think there's a small consideration with explicitly including `ParticipantCount` as a optional field in this model. IIUC, the motivation for including it here (instead of just allowing the model to ignore the extra field, which it does by default) is to validate that the value is an int (?)

I think that could make sense, but the outcome is that `ParticipantCount` becomes treated the same way as the rest of the (optional) user-provided dataset description fields, like `AccessEmail`, `AccessLink`, etc., in that the CLI will **warn** about it being unset and recommend that the user set it:

https://github.com/neurobagel/bagel-cli/blob/5a5aa88bc8c4058733c2b09602a468b037e30e91/bagel/utilities/pheno_utils.py#L673-L679

And this will happen even for an otherwise complete dataset description, where all fields are filled out besides `ParticipantCount`. If we are okay with this behaviour for the sake of ensuring `ParticipantCount` is validated when set, I think we should add this field to the [dataset description docs](https://neurobagel.org/user_guide/dataset_description/) to avoid user confusion. Otherwise, if we don't really care about validating `ParticipantCount` (i.e., we assume that it will only ever be generated by the annotation tool), I think we can just exclude it from the model entirely, since it won't be added to the output JSONLD anyways.

tests/integration/test_cli_pheno.py
```diff
@@ -162,7 +163,6 @@ def test_invalid_inputs_are_handled_gracefully(
     )
 
     assert result.exit_code != 0
-    assert len(caplog.records) == 1
```
alyssadai, line 165: Why was this change needed? Afaict the number of expected errors should not have changed, and in this test we are only capturing errors (see `propagate_errors` fixture in the function definition)

tests/integration/test_cli_pheno.py
```diff
@@ -354,7 +350,6 @@ def test_providing_non_tsv_file_raises_error(
...
```
alyssadai, line 357: Similar to above, since this PR doesn't introduce additional error logic and this test only captures errors, not sure I understand why this specific change was needed.

tests/integration/test_cli_pheno.py
```diff
@@ -195,7 +195,6 @@ def test_multiple_columns_about_single_column_variable_raises_warning(
         catch_exceptions=False,
     )
 
-    assert len(caplog.records) == 3
```
alyssadai, line 198: Following from my comment on the dataset description model, I think this is an example of how we now will have an extra warning in every test that asserts over warnings + uses a dataset description without `ParticipantCount` (which I believe is why you needed to remove these assertions). 

Is that the desired behaviour given that `ParticipantCount` is not really meant for users to fill out (for now)?

tests/integration/test_cli_pheno.py
```diff
@@ -228,7 +227,6 @@ def test_missing_bids_levels_raises_warning(
         catch_exceptions=False,
     )
 
-    assert len(caplog.records) == 1
```
alyssadai, line 231: see comment about the expected number of warnings above

tests/integration/test_cli_pheno.py
```diff
@@ -258,7 +256,6 @@ def test_bids_neurobagel_levels_mismatch_raises_warning(
...
```
alyssadai, line 261: see comment about the expected number of warnings above

tests/integration/test_cli_pheno.py
```diff
@@ -296,7 +293,6 @@ def test_unused_missing_values_raises_warning(
...
```
alyssadai, line 299: see comment about the expected number of warnings above

tests/integration/test_cli_pheno.py
```diff
@@ -978,5 +973,32 @@ def test_backup_used_with_warning_when_request_for_config_namespaces_fails(
 
     assert result.exit_code == 0
     assert temp_output_jsonld_path.exists()
-    assert len(caplog.records) == 1
```
alyssadai, line 981: see comment about the expected number of warnings above

alyssadai (commented):
Thanks @surchs for addressing this!

All changes look functional to me. Before I approve, I left one question about the desired behaviour re: treating `ParticipantCount` as an explicit optional field vs. just letting the Pydantic model ignore it. Mind taking a look?
 
Also: mind updating the PR title and adding a PR prefix?

tests/data/example24.tsv
surchs: Good catch - I think I copied them with the ISO value from a dictionary

surchs:
Thanks @alyssadai for catching this, indeed most of the changes were not needed. I have reverted them and re-requested your review

tests/integration/test_cli_pheno.py
```diff
@@ -973,32 +979,5 @@ def test_backup_used_with_warning_when_request_for_config_namespaces_fails(
 
     assert result.exit_code == 0
     assert temp_output_jsonld_path.exists()
+    assert len(caplog.records) == 1
     assert "Using a packaged backup configuration" in caplog.text
-
-
-def test_catalog_pheno_input_is_parsed(
```
alyssadai: I think in reverting the other changes this new one may have been removed by accident 😛

alyssadai (approved):
Looks good @surchs! I think one new test was removed by accident - maybe double check in case that was not intentional, then 🧑‍🍳

tests/integration/test_cli_pheno.py
surchs: Great catch!

